Issue #8, opened 2013-02-17
Remove the requirement for users to specify alignment


hgomersall:
Users are required to specify the alignment for n_byte_align and n_byte_align_empty. This is probably bad. Remove this either by using FFTW for memory management or by more direct use of pyfftw.simd_alignment.

hgomersall:
Fixed by eb5645e4a0d660b06734cf3b20c9de7afbc3bc62.

PR #44, opened 2014-07-19
close issue #8

drwells:
I added code and documentation to pick the alignment automatically.

I don't know how to write tests for this since it depends on hardware. I tried it on multiple computers and it always picked the correct alignment.

### DIFF
--- a/pyfftw/__init__.py
+++ b/pyfftw/__init__.py
@@ -22,6 +22,8 @@ from .pyfftw import (
         n_byte_align_empty,
         n_byte_align,
         is_n_byte_aligned,
+        byte_align,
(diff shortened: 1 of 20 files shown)


drwells:
I finally had time to finish modifying the tests. Let me know what you think.

hgomersall:
This is great work, thanks for it all. The tutorial is currently out of date, but that can be fixed separately (if you have time... :) - otherwise I'll get onto it when I'm not so snowed under). Apologies for the slowness of the review.

drwells:
Not a problem. I'll try to find time to work on the tutorial in the next week.

PR #54, opened 2015-01-24
Updated the tutorial.

drwells:
It no longer refers to deprecated functions like `n_byte_align`.

### DIFF
--- a/sphinx/tutorial.rst
+++ b/sphinx/tutorial.rst
@@ -47,7 +47,7 @@ caveat [#caveat]_) drop in replacements for :mod:`numpy.fft` and
 
    >>> import pyfftw
    >>> import numpy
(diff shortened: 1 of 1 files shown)


hgomersall:
That's great. Thanks for that and well done :)

drwells:
No problem. I have no idea why the python 3 build failed. Do you have any ideas?

hgomersall:
Nope. I think it was a problem in building scipy, but I'm not sure and I don't know why this would fail rather than error. I've solved it by switching to miniconda, which I made a cursory attempt to do a while ago but this time actually understood what I was doing! It actually picked up a bug in the cache code that only showed up in Python 3.4, which I was previously neglecting to test, so good work test benches :)

PR #59, opened 2015-04-06
Use complete scipy import in __init__ feature test.

bhawkins:
I was working on a Centos 6.4 machine whose package manager provides scipy 0.7.2.  This version does not have scipy.fftpack.dct (it was introduced in 0.8.0).  This leads to a situation where the try/except guard on the scipy import passes (pyfftw/interfaces/**init**.py), but actually loading the interface fails (pyfftw/interfaces/scipy_fftpack.py).  This pull request copies the actual scipy import used by the interface into the try/except block so that I can still import pyfftw if I've got an old version of scipy on my system.

### DIFF
--- a/pyfftw/interfaces/__init__.py
+++ b/pyfftw/interfaces/__init__.py
@@ -225,6 +225,9 @@ try:
 except ImportError:
     pass
 else:
-    del scipy
+    del (dct, idct, diff, tilbert, itilbert,
+        hilbert, ihilbert, cs_diff, sc_diff, ss_diff, cc_diff,
+        shift, fftshift, ifftshift, fftfreq, rfftfreq,
+        convolve, _fftpack)
     from . import scipy_fftpack
 
--- a/pyfftw/interfaces/__init__.py
+++ b/pyfftw/interfaces/__init__.py
@@ -218,7 +218,10 @@ from . import (
         cache,)
 
 try:
-    import scipy.fftpack
+    from scipy.fftpack import (dct, idct, diff, tilbert, itilbert,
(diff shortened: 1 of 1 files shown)


hgomersall:
I'm happy to commit this, but shouldn't [this line](https://github.com/bhawkins/pyFFTW/blob/90feaf1ac4afe2f81d631f5ccae897d082a818fa/test/test_pyfftw_scipy_interface.py#L152) fail? Can you add an adjustment for that...?

bhawkins:
Unfortunately I can't run the unit tests on that box because it's got python 2.6.  This means that "python -m unittest discover" doesn't work, and unittest.skip is not implemented.

hgomersall:
I presume you don't want to import scipy fftpack at all then? I misunderstood initially, I thought you wanted to just import the reduced namespace.

I think this is should be done in the same way as the existing restricted namespace. pyFFTW doesn't care about having dct.

see:
https://github.com/hgomersall/pyFFTW/blob/master/pyfftw/interfaces/scipy_fftpack.py#L68

If you do that, it would be great if you could adjust the way in which `__all__` is implemented.
The `__all__` defined on line 73 should be above the `dst` import, and then the `dst` import should append the `__all__` list.

`dct` should be done in exactly the same way.

bhawkins:
Okay thanks for the feedback. I'll take a crack at it this weekend.

hgomersall:
As per #186, I think we're going to close this. @bhawkins any last appeals?

bhawkins:
No that sounds fine. Sorry I totally forgot about this. Thanks for following up!

Issue #13, opened 2013-04-18
Allow GIL to be released with threads=1


hgomersall:
Currently the GIL is never released during executions if the FFTW object is single threaded. In a Python multithreaded environment, it may be helpful to execute FFTW objects in parallel in different Python threads so there should be a way for the user to turn off the GIL even when the number of threads used by the FFTW object is 1.

r-atwood:
Hi, 
is this an issue for the following use, many instances of the same FFT being called on each of a 
large number of identical arrays. The routine uses the multiprocessing module . Abbreviated code is below. 

In this use, I have two questions that seem to be related to this issue , 

1. Does this suffer from not ulocking the GIL as mentioned, thus it should be faster than it is? It seems faster than using one single thread already. 

2. Is it threadsafe against multiple threads modifying the array being processed in fft_object? I thought that fftw used the exact array that was designated at "planning" time ? Or is it always using an array of identical type but in a different memory location? 

Thanks 
Robert

____________________________________________________
import pyfftw
import multiprocessing.

**** In main routine ***  

fft_object = pyfftw.FFTW(a, b, axes=(0,1))
ifft_object = pyfftw.FFTW(c, d, axes=(0,1), direction='FFTW_BACKWARD')


for i in range(n):
     p = multiprocessing.Process(target=ravenfilter,args=((..blah..blah.., fft_object,ifft_object)))
     jobs.append(p)
     p.start()

**** In a function****
def ravenfilter( ... several args... , fft_object,ifft_object):
     ... get the array 'sinogram' somehow ... 
      sino2=ft_object(sinogram)
     ... more operations ..

hgomersall:
1) Lots of calls to identical arrays should not really be affected by the GIL issue unless you're running stuff in multiple threads yourself. Multiprocessing uses different processes anyway so shouldn't really be affected either way.

It's perfectly plausible that many small transforms are faster in a single thread in which the threading overhead is greater than any speed up. As always, try it and see :)

2) If multiple threads hit the same array, you should expect things to break. The idea is that other threads might be doing other things (or running other FFTW instances or whatever) and so the GIL can be released whilst FFTW is working.

Does that answer the qs?

r-atwood:
Thanks for the speedy reply.
I am afraid it may answer the question by referring to things I am not familiar enough with, in Python. 
If you can clarify a couple of things about your answer it would be very helpful. 
[...]

PR #77, opened 2016-01-14
Release GIL during single- and multi-thread execution

zpincus:
See issue #13.

Also removes code made redundant by no longer needing to decide whether or not to use nogil based on thread count.

### DIFF
--- a/pyfftw/pyfftw.pyx
+++ b/pyfftw/pyfftw.pyx
@@ -1540,7 +1540,10 @@ cdef class FFTW:
         cdef void *output_pointer = (
                 <void *>np.PyArray_DATA(self._output_array))
         
-        self._fftw_execute(self._plan, input_pointer, output_pointer)
+        cdef void *plan = self._plan
+        cdef fftw_generic_execute fftw_execute = self._fftw_execute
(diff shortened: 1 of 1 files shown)


hgomersall:
Looks good. I'm not aware of any other places in the docs where this issue is referenced (and a cursory search would agree with that).

Assuming all the tests pass, I'll merge.

zpincus:
Hold off on merging this... more detailed testing shows that my changes prevent the GIL from ever being released. It might be something to do with how cython handles nogil function pointers (or fails to). I will do some research and push some new changes. Sorry!!

zpincus:
OK, my bad. I had misread the cython docs as saying that the `nogil` annotation implicitly releases the gil when the function is called. This is not the case: it just makes it not an error to call that function in a `with nogil` block. Will fix ASAP.

zpincus:
The PR is all better now. Sorry for the idiocy.

hgomersall:
Ah, well done. I suppose I should have picked that up!

zpincus:
Closing this PR to provide a new one with cleaner history.

Issue #75, opened 2016-01-14
Release GIL during planning?


zpincus:
Currently, while calling into FFTW to create a plan, pyFFTW holds onto the GIL. So it's not possible to have a second python thread executing while FFTW does its planning. (Use case: I want to log a warning if planning takes more than 5 seconds or whatever, so that I know if something went wonky with my saved wisdom file. Thus, firing off a timer in another thread seems like the thing to do, except for the GIL not being released by pyFFTW, which means my second python thread doesn't run until the planning is done.)

Would adding `with nogil:` around [pyfftw.pyx:1086](https://github.com/hgomersall/pyFFTW/blob/master/pyfftw/pyfftw.pyx#L1086)
be a safe and sane change? Alternately the plan functions could be declared [nogil](http://docs.cython.org/src/userguide/external_C_code.html#declaring-a-function-as-callable-without-the-gil).

This is not unrelated to issue #13, for that matter. There are reasonable use-cases where one would want to release the GIL during single-threaded FFTW execution so as to run something else in a different python thread concurrently. Is there any reason the code refuses to release the GIL during single-threaded FFTW execution? 

As an aside, the [documentation](https://hgomersall.github.io/pyFFTW/pyfftw/pyfftw.html) states that "If the number of threads is greater than 1, then the GIL is released by necessity". I'm not sure this is actually the case, and this may indicate some misconceptions about the GIL. C code that spins up its own threads would still work fine if the GIL is held, unless those threads need to interact with the python interpreter in any way (such as calling python callbacks). For FFTW this seems to not be the case, right?

The reason to release the GIL doesn't have to do with whether or not the called code is multithreaded -- releasing the GIL is strictly a nicety to allow other python threads to run whilst *any* C code executes  (single- or multi-threaded). Otherwise one python thread's long-running call into C will block *all* python threads. So, unless I am rather mistaken (a distinct possibility!), the number of threads FFTW is going to use should have nothing to do with whether pyFFTW should release the GIL before calling into FFTW... 

Put another way, I don't see any reason why every FFTW function that does non-trivial computation shouldn't be declared `nogil` in the cython files. Or at lease all functions that don't need python callbacks (if there are any that do). Am I missing something? Happy to get a PR together if it would help.

Thanks!

hgomersall:
Yeah, very good point, so thanks.

[...]

PR #78, opened 2016-01-14
Release GIL around FFTW planning

zpincus:
Instead acquire a python-level lock to keep multiple threads from planning simultaneously. Addresses Issue #75.

### DIFF
--- a/pyfftw/pyfftw.pyx
+++ b/pyfftw/pyfftw.pyx
@@ -1101,13 +1101,11 @@ cdef class FFTW:
         cdef void *_in = <void *>np.PyArray_DATA(self._input_array)
         cdef void *_out = <void *>np.PyArray_DATA(self._output_array)
         cdef int sign = self._direction
-        cdef int c_flags = self._flags
-        
-        plan_lock.acquire()
-        with nogil:
-            plan = fftw_planner(rank, dims, howmany_rank, howmany_dims, 
-                _in, _out, sign, c_flags)
-        plan_lock.release()
+        cdef unsigned c_flags = self._flags
+
+        with plan_lock, nogil:
+            plan = fftw_planner(rank, dims, howmany_rank, howmany_dims,
+                                _in, _out, sign, c_flags)
         self._plan = plan
         
         if self._plan == NULL:
--- a/pyfftw/pyfftw.pyx
+++ b/pyfftw/pyfftw.pyx
@@ -39,6 +39,7 @@ from libc.stdint cimport intptr_t, int64_t
 from libc cimport limits
 
 import warnings
+import threading
 
 include 'utils.pxi'
 
@@ -63,6 +64,10 @@ flag_dict = {'FFTW_MEASURE': FFTW_MEASURE,
 
 _flag_dict = flag_dict.copy()
 
+# Need a global lock to protect FFTW planning so that multiple Python threads
+# do not attempt to plan simultaneously.
+PLAN_LOCK = threading.Lock()
+
 # Function wrappers
 # =================
 # All of these have the same signature as the fftw_generic functions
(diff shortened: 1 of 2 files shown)


pyfftw/pyfftw.pyx
```diff
@@ -1082,14 +1087,29 @@ cdef class FFTW:
...
+        cdef fftw_iodim *howmany_dims = <fftw_iodim *>self._howmany_dims
+        cdef void *_in = <void *>np.PyArray_DATA(self._input_array)
+        cdef void *_out = <void *>np.PyArray_DATA(self._output_array)
+        cdef int sign = self._direction
+        cdef int c_flags = self._flags
+        
+        plan_lock.acquire()
+        with nogil:
+            plan = fftw_planner(rank, dims, howmany_rank, howmany_dims, 
+                _in, _out, sign, c_flags)
```
hgomersall: It's a weak requirement, but can we make this more pep8y (line up the arguments, unless all on new line).

pyfftw/pyfftw.pyx
```diff
@@ -1082,14 +1087,29 @@ cdef class FFTW:
...
+        plan_lock.release()
+        self._plan = plan
+        
         if self._plan == NULL:
             raise RuntimeError('The data has an uncaught error that led '+
                     'to the planner returning NULL. This is a bug.')
```
hgomersall: This is also not pep8 compliant! (not your problem - legacy code...)

pyfftw/pyfftw.pyx
hgomersall: Actually, don't bother with this. I'll go through at some point and make everything pep8 compliant.

pyfftw/pyfftw.pyx
```diff
@@ -1082,14 +1087,29 @@ cdef class FFTW:
...
+        cdef int rank = self._rank
+        cdef fftw_iodim *dims = <fftw_iodim *>self._dims
+        cdef int howmany_rank = self._howmany_rank
+        cdef fftw_iodim *howmany_dims = <fftw_iodim *>self._howmany_dims
+        cdef void *_in = <void *>np.PyArray_DATA(self._input_array)
+        cdef void *_out = <void *>np.PyArray_DATA(self._output_array)
+        cdef int sign = self._direction
+        cdef int c_flags = self._flags
+        
+        plan_lock.acquire()
```
hgomersall: Is there a reason this can't be done with a context manager?

zpincus:
OK, merged in the new master and fixed things to be more pep-8-y and use a context manager for the lock. (I didn't realize you could mix a cython "with nogil" block with normal python context managers, but cython handles this quite nicely.)

Let me know if you'd again prefer a `merge --squash` branch once this is all as desired.

hgomersall:
No, no probs - history is not a problem per se, I just don't like all the crap history that the CI systems seem to result in.

zpincus:
I have no idea why the last commit broke the appveyor build! The previous commit tested fine, but the latest one (which literally only changed a single comment!) is failing on some but not all platforms with:
```
======================================================================
FAIL: test_objects_removed_after_keepalive (test.test_pyfftw_interfaces_cache.CacheTest)
----------------------------------------------------------------------
  File "C:\projects\pyfftw\test\test_pyfftw_interfaces_cache.py", line 283, in test_objects_removed_after_keepalive
AssertionError: KeyError not raised by lookup
----------------------------------------------------------------------
```

Any ideas what's going on here?

hgomersall:
Yes, the issue I think is that appveyor builds are scheduled very little time on a virtual machine. The tests that fail depend on a timing threshold not being passed, which it does, because it's just not running to do what it needs to do.

The test should prob be fixed to use cpu time rather than wall time. Though it is actually wall time that is measured.

zpincus:
Ah, OK. Glad to hear my patches didn't somehow cause this!

Let me know if you think this patch-set needs anything else (threaded test-case to demonstrate that the GIL is released during planning? That might be a bit tricky but probably do-able if you think it worthwhile.) or if it looks good to go.

hgomersall:
No, I've no idea how to reliably test threaded release (unless you have any bright ideas). The current failing highlights the problem with testing what is essentially a designed in race condition.

zpincus:
No real bright ideas either that don't basically rely on race conditions. (Actually, any such test would almost by definition be a race condition.)

PR #79, opened 2016-01-14
Add support for FFTW_WISDOM_ONLY

zpincus:
Add support and tests for the FFTW_WISDOM_ONLY planning flag, which causes errors to be raised if the plan is not in the wisdom (instead of actually doing the possibly time-consuming planning).

Use case: to detect if a plan is not in the wisdom, and if so generate a warning or take another course of action before performing a potentially slow planning operation. (Most useful with large transforms using FFTW_PATIENT or FFTW_EXHAUSTIVE, where planning can take tens of minutes to hours...)


hgomersall:
hmmm, it seems to fail on python 3.5 under appveyor...

zpincus:
Huh... it's failing on 32-bit platforms (all python versions). I am not sure what to make of this -- some difference in FFTW? Could the `FFTW_WISDOM_ONLY` constant (2097152) be different on 32-bit systems? Doesn't seem to be the case from `fftw.h`? I'll try to get a 32-bit build working on my machine, but I'm not sure exactly how to debug this!

zpincus:
OK, at this point I have no idea what's going on here. I fixed a few potential issues (use of signed vs. unsigned ints), but so far no luck getting the appveyor check to work.

Worse, I built a 32-bit python / FFTW on my mac and cannot replicate the problem. Moreover, the errors seem somewhat random on the 32-bit appveyor builds -- it's not consistent python version to python version, or between test runs. So maybe there's some memory that's getting overwritten or something else non-deterministic happening?

This is more than a little frustrating since my changes are so simple, and because I can't locally replicate the error. So at this point I can't think of anything better to do than to try changes that seem perhaps helpful and see if that fixes things on the appveyor platform.

test/test_pyfftw_complex.py
```diff
@@ -188,6 +190,31 @@ def test_planner_flags(self):
...
+        # run_validate_fft copies the input arrays, which can randomly
+        # cause them to be aligned or unaligned. Without forcing unaligned, this
+        # could otherwise make the wisdom invalid for future planning...
+        self.assertRaisesRegex(RuntimeError, 'No FFTW wisdom', 
+                self.run_validate_fft, *(a, b, axes), 
+                **{'flags':('FFTW_ESTIMATE', 'FFTW_WISDOM_ONLY'), 
+                   'force_unaligned_data': True})
+        # now plan the FFT
+        self.run_validate_fft(a, b, axes, flags=('FFTW_ESTIMATE',),
+                force_unaligned_data=True)
```
hgomersall: I'm not sure how FFTW handles the FFTW unaligned flag. It's not clear to me that it doesn't still need the data to still be on the correct alignment even if it doesn't use it to construct the relevant plan. I know that the planner is very conservative about what it allows.

hgomersall:
I think it's worth changing the tests to use aligned arrays. If that works, I think we can accept the patch - the problem of FFTW being overly fussy about the plans is not worth spending too much time understanding (it's going to be a fairly niche flag anyway).

It might be worth dropping Matteo Frigo an email to understand better why `FFTW_UNALIGNED` plans are not compatible with one another.

hgomersall:
The other thing to try would be to print out the wisdom (it's just a string) in the two cases when it fails.

I'll try to look at this on windows at some point - Appveyor builds a wheel as part of the process (which it uploads on success), so that can be grabbed easily enough by uncommenting line 80 of `appveyor.yml` and connecting to the virtual machine over remote desktop.

hgomersall:
Just to add a bit more to this. The `FFTW_UNALIGNED` flag documentation says the following:

> `FFTW_UNALIGNED` specifies that the algorithm may not impose any unusual alignment requirements on the input/output arrays (i.e. no SIMD may be used). This flag is normally not necessary, since the planner automatically detects misaligned arrays. The only use for this flag is if you want to use the new-array execute interface to execute a given plan on a different array that may not be aligned like the original. (Using fftw_malloc makes this flag unnecessary even then. You can also use fftw_alignment_of to detect whether two arrays are equivalently aligned.) 

It actually doesn't make any claims about the wisdom that is created when the `FFTW_UNALIGNED` flag is used, only that a plan can be used with any alignment.

zpincus:
Thanks for your comments and help -- I have finally fixed things so that the tests work properly! Much more fuss than I intended for such a minor feature.

I have verified that alignment still matters to the wisdom, even if FFTW_UNALIGNED is specified. (Similarly, wisdom for an aligned array generated using FFTW_UNALIGNED can't be used for the same array without FFTW_UNALIGNED.) I'm not sure why I thought FFTW_UNALIGNED would be a good solution to these test problems... The latest commit does the obvious right thing, which is to make the test routines not copy the arrays in the first place.

In any case, if you want to merge this set of changes but prefer the history to look nicer, I can make a new topic branch and rebase to include only the final set of changes. (Right now as you see there are a few commits to test certain ideas on Appveyor and then revert those changes...) A PR from a rebased branch will look a little simpler.

hgomersall:
It would be slightly nicer to have a rebased branch, but don't worry _too_ much. When I'm messing with getting the CI systems working, I tend to branch specifically for that purpose, fiddle a bit, and then `merge --squash` back in (removing all the intermediate broken CI config files).

hgomersall:
I like the test now, it's clear and precise. Thanks for your work on this.

zpincus:
Thanks again for your help on all of this! I know it's a super-niche feature, but this plus the threading PRs will be super-valuable for my lab's FFTW uses.

Also, I didn't know about `merge --squash` (easier to make work correctly than rebasing for sure!)

Anyhow, I have closed this PR and opened a new one for my `wisdom-only-final` branch, which contains the squashed merge.

PR #80, opened 2016-01-18
Support FFTW_WISDOM_ONLY (cleaned up patch)

zpincus:
Add support and tests for the FFTW_WISDOM_ONLY planning flag, and also address potential type incompatibilities between FFTW.h and pyfftw (int vs. unsigned for the FFTW flags parameter).


hgomersall:
Great, looks good. When the tests have completed, I'll merge.

Issue #92, opened 2016-03-27
Numpy interface can fail with non-writeable arrays.


mforbes:
The following code fails if it is called before a plan is cached:

```python
import numpy as np
from pyfftw.interfaces.numpy_fft import fat
y = np.zeros(4) + 1j  # Needs to be complex for failure
y.flags.writeable = False
fft(y)
```

```
In [5]: fft(y)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
...
.../pyfftw/interfaces/_utils.pyc in _Xfftn(a, s, axes, overwrite_input, planner_effort, threads, auto_align_input, auto_contiguous, calling_func)
     81         a_copy = a.copy()
     82 
---> 83         FFTW_object = getattr(builders, calling_func)(*args)
     84 
     85         # Only copy if the input array is what was actually used
...
.../pyfftw/builders/_utils.pyc in _Xfftn(a, s, axes, overwrite_input, planner_effort, threads, auto_align_input, auto_contiguous, avoid_copy, inverse, real)
    171         if not avoid_copy:
    172             # Copy the data back into the (likely) destroyed array
--> 173             FFTW_object.get_input_array()[:] = a_copy
    174 
    175     return FFTW_object

ValueError: assignment destination is read-only
```

The problem is that the plan is made using the original array, which in this case is not writeable.  Since a copy is made, why not use it for planning instead since the copy will always be writeable?

The minimal change would be to set and restore the writeable flag:

```patch
82a83,85
>         # Input array must be writeable...
>         writeable, a.flags.writeable = a.flags.writeable, True
> 
89a93,95
>         # Restore writeable flag
>         a.flags.writeable = writeable
> 
```

Using the copy instead seems to work but there may be subtleties I am not aware of.

```patch
24c24
< 
---
> import numpy as np
81c81
<         a_copy = a.copy()
---
>         args = (np.empty_like(a),) + args[1:]
85,89d84
<         # Only copy if the input array is what was actually used
<         # (otherwise it shouldn't be overwritten)
<         if FFTW_object.get_input_array() is a:
<             a[:] = a_copy
< 
93c88
<         output_array = FFTW_object()
---
>         output_array = FFTW_object(input_array=a) 
```

hgomersall:
This is an interesting one. I think the only safe behaviour is to not touch an array that is marked read only, in which case, we should basically require, at least for the numpy interface, that the input have its writeable flag as True. This is definitely a bug, so thanks.

[...]

PR #93, opened 2016-03-29
Resolves issue #92.

mforbes:
The essence of this change is to use copies of the input array `a` for planning purposes or when the array must be restored after a transform.  This prevents accessing the input array as much as possible, allowing it to be locked, and thereby resolving issue #92 in all cases except when `overwrite_input == True`.  One might like to raise an error if this is the case while `a.flags.writeable == True` but I have kept the changes minimal in this PR.

### DIFF
--- a/pyfftw/interfaces/_utils.py
+++ b/pyfftw/interfaces/_utils.py
@@ -5,6 +5,9 @@
 # Henry Gomersall
 # heng@kedevelopments.co.uk
 #
+# Michael McNeil Forbes
+# michael.forbes+pyfftw@gmail.com
+#
 # All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
@@ -47,7 +50,7 @@ def _Xfftn(a, s, axes, overwrite_input, planner_effort,
         threads, auto_align_input, auto_contiguous, 
         calling_func, normalise_idft=True):
 
-    reload_after_transform = False
+    work_with_copy = False
 
     a = numpy.asanyarray(a)
 
@@ -63,21 +66,35 @@ def _Xfftn(a, s, axes, overwrite_input, planner_effort,
 
     if calling_func in ('irfft2', 'irfftn'):
         # overwrite_input is not an argument to irfft2 or irfftn
-        args = (a, s, axes, planner_effort, threads, 
-                auto_align_input, auto_contiguous)
+        args = (planner_effort, threads, auto_align_input, auto_contiguous)
 
         if not overwrite_input:
             # Only irfft2 and irfftn have overwriting the input
             # as the default (and so require the input array to 
             # be reloaded).
-            reload_after_transform = True
+            work_with_copy = True
     else:
-        args = (a, s, axes, overwrite_input, planner_effort, threads, 
+        args = (overwrite_input, planner_effort, threads,
                 auto_align_input, auto_contiguous)
     
+        if not a.flags.writeable:
+            # Special case of a locked array - always work with a
+            # copy.  See issue #92.
+            work_with_copy = True
+
+            if overwrite_input:
+                raise ValueError(
+                    "Cannot set overwrite_input when not a.flags.writeable")
+
+    if work_with_copy:
+        # We make the copy before registering the key so that the
+        # copy's stride information will be cached since this will be
+        # used for planning.
+        a = a.copy()
+
     if cache.is_enabled():
         key = (calling_func, a.shape, a.strides, a.dtype, s.__hash__(), 
-                axes.__hash__(), args[3:])
+               axes.__hash__(), args)
 
         try:
             if key in cache._fftw_cache:
@@ -92,18 +109,21 @@ def _Xfftn(a, s, axes, overwrite_input, planner_effort,
 
     if not cache.is_enabled() or FFTW_object is None:
 
-        # If we're going to create a new FFTW object, we need to copy
-        # the input array to preserve it, otherwise we can't actually
-        # take the transform of the input array! (in general, we have
-        # to assume that the input array will be destroyed during 
-        # planning).
-        a_copy = a.copy()
+        # If we're going to create a new FFTW object and are not
+        # working with a copy, then we need to copy the input array to
+        # preserve it, otherwise we can't actually  take the transform
+        # of the input array! (in general, we have to assume that the
+        # input array will be destroyed during planning).
+        if not work_with_copy:
+            a_copy = a.copy()
 
-        FFTW_object = getattr(builders, calling_func)(*args)
+        planner_args = (a, s, axes) + args
+
+        FFTW_object = getattr(builders, calling_func)(*planner_args)
     
         # Only copy if the input array is what was actually used
         # (otherwise it shouldn't be overwritten)
-        if FFTW_object.input_array is a:
+        if not work_with_copy and FFTW_object.input_array is a:
             a[:] = a_copy
 
         if cache.is_enabled():
@@ -112,9 +132,6 @@ def _Xfftn(a, s, axes, overwrite_input, planner_effort,
         output_array = FFTW_object(normalise_idft=normalise_idft)
 
     else:
-        if reload_after_transform:
-            a_copy = a.copy()
-
         orig_output_array = FFTW_object.output_array
         output_shape = orig_output_array.shape
         output_dtype = orig_output_array.dtype
@@ -126,7 +143,4 @@ def _Xfftn(a, s, axes, overwrite_input, planner_effort,
         FFTW_object(input_array=a, output_array=output_array, 
                 normalise_idft=normalise_idft)
     
-    if reload_after_transform:
-        a[:] = a_copy
-
     return output_array
--- a/pyfftw/interfaces/_utils.py
+++ b/pyfftw/interfaces/_utils.py
@@ -89,8 +89,11 @@ def _Xfftn(a, s, axes, overwrite_input, planner_effort,
     if work_with_copy:
         # We make the copy before registering the key so that the
         # copy's stride information will be cached since this will be
-        # used for planning.
-        a = a.copy()
+        # used for planning.  Make sure the copy is byte aligned to
+        # prevent further copying
+        a_original = a
+        a = pyfftw.empty_aligned(shape=a.shape, dtype=a.dtype)
+        a[...] = a_original
 
     if cache.is_enabled():
         key = (calling_func, a.shape, a.strides, a.dtype, s.__hash__(), 
--- a/pyfftw/interfaces/_utils.py
+++ b/pyfftw/interfaces/_utils.py
@@ -83,8 +83,8 @@ def _Xfftn(a, s, axes, overwrite_input, planner_effort,
             work_with_copy = True
 
             if overwrite_input:
-                raise ValueError(
-                    "Cannot set overwrite_input when not a.flags.writeable")
+                raise ValueError('overwrite_input cannot be True when the ' +
+                                 'input array flags.writeable is False')
 
     if work_with_copy:
         # We make the copy before registering the key so that the
--- a/test/test_pyfftw_numpy_interface.py
+++ b/test/test_pyfftw_numpy_interface.py
@@ -593,7 +593,8 @@ class InterfacesNumpyFFTTestFFT(unittest.TestCase):
 
         self.assertRaisesRegex(
             ValueError,
-            'Cannot set overwrite_input when not a.flags.writeable',
+            'overwrite_input cannot be True when the ' +
+            'input array flags.writeable is False',
             interfaces.numpy_fft.fft,
             a, overwrite_input=True)
 
--- a/test/test_pyfftw_numpy_interface.py
+++ b/test/test_pyfftw_numpy_interface.py
@@ -582,19 +582,11 @@ class InterfacesNumpyFFTTestFFT(unittest.TestCase):
                               test_shape, dtype, s, kwargs,
                               copy_func=copy_with_writeable)
 
-    def test_regression_issue_92(self):
-        '''Quick regression test for issue 92.
-
-        The previous unit tests do not cover this when quick_tests is
-        run because one needs the complex dtype for failure.
+    def test_overwrite_input_for_issue_92(self):
+        '''Tests that trying to overwrite a locked array fails.
         '''
-        numpy.random.seed(1)
-        a = numpy.random.random((100,)) + 1j
-        interfaces.numpy_fft.fft(a)
-        a = numpy.random.random((4, 4)) + 1j
+        a = numpy.zeros((4,))
         a.flags.writeable = False
-        assert numpy.allclose(np_fft.fft(a), interfaces.numpy_fft.fft(a))
-
         self.assertRaisesRegex(
             ValueError,
             'overwrite_input cannot be True when the ' +
--- a/test/test_pyfftw_numpy_interface.py
+++ b/test/test_pyfftw_numpy_interface.py
@@ -591,6 +591,12 @@ class InterfacesNumpyFFTTestFFT(unittest.TestCase):
         a.flags.writeable = False
         assert numpy.allclose(np_fft.fft(a), interfaces.numpy_fft.fft(a))
 
+        self.assertRaisesRegex(
+            ValueError,
+            'Cannot set overwrite_input when not a.flags.writeable',
+            interfaces.numpy_fft.fft,
+            a, overwrite_input=True)
+
 
 class InterfacesNumpyFFTTestIFFT(InterfacesNumpyFFTTestFFT):
     func = 'ifft'
--- a/test/test_pyfftw_numpy_interface.py
+++ b/test/test_pyfftw_numpy_interface.py
@@ -148,24 +148,25 @@ class InterfacesNumpyFFTTestFFT(unittest.TestCase):
             self.assertRaisesRegex = self.assertRaisesRegexp
 
     def validate(self, array_type, test_shape, dtype, 
-            s, kwargs):
+                 s, kwargs, copy_func=copy.copy):
 
         # Do it without the cache
 
         # without:
         interfaces.cache.disable()
-        self._validate(array_type, test_shape, dtype, s, kwargs)
+        self._validate(array_type, test_shape, dtype, s, kwargs,
+                       copy_func=copy_func)
 
     def munge_input_array(self, array, kwargs):
         return array
 
     def _validate(self, array_type, test_shape, dtype, 
-            s, kwargs):
+                  s, kwargs, copy_func=copy.copy):
 
         input_array = self.munge_input_array(
                 array_type(test_shape, dtype), kwargs)
 
-        orig_input_array = copy.copy(input_array)
+        orig_input_array = copy_func(input_array)
 
         np_input_array = numpy.asarray(input_array)
         
@@ -189,13 +190,13 @@ class InterfacesNumpyFFTTestFFT(unittest.TestCase):
 
             try:
                 test_out_array = getattr(self.validator_module, self.func)(
-                        copy.copy(np_input_array), s, **axes)
+                        copy_func(np_input_array), s, **axes)
 
             except Exception as e:
                 interface_exception = None
                 try:
                     getattr(self.test_interface, self.func)(
-                            copy.copy(input_array), s, **kwargs)
+                            copy_func(input_array), s, **kwargs)
                 except Exception as _interface_exception:
                     # It's necessary to assign the exception to the
                     # already defined variable in Python 3.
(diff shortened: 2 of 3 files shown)


pyfftw/interfaces/_utils.py
```diff
@@ -97,24 +101,25 @@ def _Xfftn(a, s, axes, overwrite_input, planner_effort,
         # take the transform of the input array! (in general, we have
         # to assume that the input array will be destroyed during 
         # planning).
-        a_copy = a.copy()
+        if overwrite_input:    # Note: reload_after_transform is not True...
+            args = (a, s, axes) + args
+            a_copy = a.copy()  # ... hence this is not an additional copy
+        else:
+            args = (numpy.empty_like(a), s, axes) + args
```
hgomersall: `empty_like` doesn't create an array with the correct strides etc. It should prob be called `empty_sorta_like`.

pyfftw/interfaces/_utils.py
```diff
@@ -126,7 +131,4 @@ def _Xfftn(a, s, axes, overwrite_input, planner_effort,
         FFTW_object(input_array=a, output_array=output_array, 
                 normalise_idft=normalise_idft)
```
hgomersall: Two of the transform types will always clobber the input, so it's not safe to drop the reload after input for those.

hgomersall:
I made a couple of notes to the diff - also to the original side.

pyfftw/interfaces/_utils.py
mforbes: In these cases, the transform is performed on a copy of the original array, so the original would not be clobbered.

pyfftw/interfaces/_utils.py
```diff
@@ -90,6 +89,11 @@ def _Xfftn(a, s, axes, overwrite_input, planner_effort,
             # the check and the lookup
             FFTW_object = None
 
+    if reload_after_transform:
+        # Instead of reloading after transform, we just transform a
+        # copy instead.
+        a = a.copy(order='K')  # Match original layout as closely as possible
```
mforbes: Note that in the two special cases where FFTW will always clobber the original array, we instead transform a copy.  In this particular case, the copy will also be used for planning purposes so there is no issue with striding.

hgomersall: Ah ok, that makes sense.

pyfftw/interfaces/_utils.py
mforbes: This is the main potential issue and does not seem to be covered in the unit tests.  Can you give me an example?  I tried the following, but it seems to work fine.

```python
import numpy as np
from pyfftw.interfaces import numpy_fft as npfft
A = np.random.random((10,10))
B = A[0::2, 0::2]  # Non-contiguous input array.
np.allclose(np.fft.fftn(B), npfft.fftn(B))
```

pyfftw/interfaces/_utils.py
```diff
@@ -63,8 +63,7 @@ def _Xfftn(a, s, axes, overwrite_input, planner_effort,
 
     if calling_func in ('irfft2', 'irfftn'):
         # overwrite_input is not an argument to irfft2 or irfftn
-        args = (a, s, axes, planner_effort, threads, 
-                auto_align_input, auto_contiguous)
+        args = (planner_effort, threads, auto_align_input, auto_contiguous)
```
hgomersall: This prob means the key on line 79 should be updated as well - `args[3:]` -> `args`.

mforbes: Yes.  I thought I did that but apparently it did not make it into the git version.

hgomersall:
Apologies, I misinterpreted what you were doing. This seems like a neat way to solve the problem (and frankly, should have been how it was originally implemented).

It would be great to have an accompanying test case for the `writeable=False` situation (which should fail before and pass after). The easiest way would be to replicate something like [the non-numpy case](https://github.com/pyFFTW/pyFFTW/blob/master/test/test_pyfftw_numpy_interface.py#L318), but change the lambda (or create a new function) that changes the `writeable` flag.

hgomersall:
(and thanks!)

pyfftw/interfaces/_utils.py
hgomersall: `A.strides == B.strides` returns `False`. The FFTW object that is created for A (or B), cannot then be used to process the other.

It's probably possible to create an array with the same strides using `numpy.lib.stride_tricks.as_strided`, but it would need a bit of thinking about to work robustly.

pyfftw/interfaces/_utils.py
mforbes: This is fixed now.

pyfftw/interfaces/_utils.py
mforbes: I know that the strides are different, but the `fftn` call still works even though the plan is created on an array with different strides than it is executed on.  Can you find a case that fails?  (I would really be interested in this because I have always worried about planning on one array and executing the transform on another, but have never had a problem.  Perhaps it is just a performance issue: the optimal plan is not generated?  That might be pretty tricky to check unless there is a way to inspect the plans...

mforbes:
It will take me a bit, but I will try to write some tests: thanks for pointing out a place to start.  If you could try to find some case in which planning with an identically strided array is important, I could add a test for that too, but I can't seem to figure out how to create a problem with the current implementation.

pyfftw/interfaces/_utils.py
```diff
@@ -97,24 +101,25 @@ def _Xfftn(a, s, axes, overwrite_input, planner_effort,
         # take the transform of the input array! (in general, we have
         # to assume that the input array will be destroyed during 
         # planning).
-        a_copy = a.copy()
+        if overwrite_input:    # Note: reload_after_transform is not True...
+            planner_args = (a, s, axes) + args
+            a_copy = a.copy()  # ... hence this is not an additional copy
+        else:
+            planner_args = (numpy.empty_like(a), s, axes) + args
```
mforbes: This should possibly use `numpy.lib.stride_tricks.as_strided` but one needs to be careful with this to make sure that it is applied to a large-enough chunk of memory.  One possibility might be to inspect `a` and make copy of the original array, then apply the `as_strided`.  However, this could cause problems if the original array is huge since `a` might be a small view.  In this case, the original code - transforming on a copy and then restoring it - would be preferable and might require one more special case.

That being said, I still can't find a way that this code actually fails.  Perhaps it is just a performance issue.

pyfftw/interfaces/_utils.py
hgomersall: It doesn't fail, it just has to plan again, so you lose the advantage of the cache (which can be quite significant) and also will potentially clobber the array (and will do so silently).

mforbes: So if you call an `FFTW_object` with an incorrectly strided array, it will execute another plan?  (clobbering the original array should be easy to test for.)

I will think about the simplest logic that allows one to work with the original array if it is indeed strided.

pyfftw/interfaces/_utils.py
hgomersall: (as a quick aside, you shouldn't plan on an empty array - processors often take different paths on encountering rare float types, which are less rare on random data, so the plan is essentially wrong)

pyfftw/interfaces/_utils.py
mforbes: Oh, I see.  You probably mean that the cache would not work, but I tricked the cache into saving the original strides:-)  However, there might be a problem (at least from a performance standpoint) applying a plan done on a strided array to an unstudied array, so I should still try to find a fix.

pyfftw/interfaces/_utils.py
mforbes: I see.  Is `zero_like()` sufficient, or should it contain "real" data?

hgomersall: Yes, I think `zero_like()` is fine.

pyfftw/interfaces/_utils.py
hgomersall: `[update_arrays](https://github.com/mforbes/pyFFTW/blob/patch-1/pyfftw/pyfftw.pyx#L1446)`, which is used by `FFTW.__call__` on line 120 of `_utils.py` is very conservative and will reject erroneous arrays (i.e. arrays different to what the FFTW object planned for) with a `ValueError`.

mforbes: Can you give me an example that uses the current code which triggers this error?

mforbes:
So it seems there is one corner case: a locked strided array with large strides so that trying to make a properly strided copy would be prohibitive.  What is best:

1. Use a potentially inefficient plan that was developed on an unstrided array? (My choice and current patch.) (Your latest comment suggests that this could trigger a `ValueError` from `update_arrays`)
2. Always plan and perform such transforms on an unstrided copy.  (Slightly more complicated logic, but similar to what is done in this patch for the `rfftn` special case).  This seems the safest in light of your recent comment.
3. Raise an exception that the cases is not supported.  (Current behaviour.)

pyfftw/interfaces/_utils.py
hgomersall: Ah, I was slightly wrong. [`FFTW.__call__`](https://github.com/mforbes/pyFFTW/blob/patch-1/pyfftw/pyfftw.pyx#L1339) will force the array to be correct before calling `update_arrays`, which means an extra copy. The problem is there, but is a performance related one as you pointed out.

mforbes:
Okay.  I simplified the logic.  Now there is a flag `work_with_copy` that is set for the two special cases of the `irfftn` and locked arrays.  In these cases the plan and the execution will always take case with a copy (which is made before the cache is checked so that the appropriate stride information is always considered in the cache.)

pyfftw/interfaces/_utils.py
```diff
@@ -63,21 +63,31 @@ def _Xfftn(a, s, axes, overwrite_input, planner_effort,
...
+        if not a.flags.writeable:
+            # Special case of a locked array - always work with a
+            # copy.  See issue #92.
+            work_with_copy = True
+
+    if work_with_copy:
+        # We make the copy before registering the key so that the
+        # copy's stride information will be cached since this will be
+        # used for planning.
+        a = a.copy(order='K')  # Match original layout as closely as possible
```
mforbes: Question: is it better to match the original calling order (C or F) or since we always work with the copy created here, would it be best to make the copy in C order for performance?

hgomersall: I don't know. Perhaps `order='A'`. It's plausible the user knows something about the array, so always forcing to be C is the wrong thing. I don't actually know what 'K' does!

mforbes: `order='K'` tries to match the input array as closely as possible.

pyfftw/interfaces/_utils.py
```diff
@@ -63,21 +63,31 @@ def _Xfftn(a, s, axes, overwrite_input, planner_effort,
...
+            work_with_copy = True
     else:
-        args = (a, s, axes, overwrite_input, planner_effort, threads, 
+        args = (overwrite_input, planner_effort, threads, 
                 auto_align_input, auto_contiguous)
     
+        if not a.flags.writeable:
+            # Special case of a locked array - always work with a
+            # copy.  See issue #92.
+            work_with_copy = True
```
hgomersall: We need to be a bit more careful with this. There is a potential conflict between `writeable` and `overwrite_input`. If `writeable` is `False`, `overwrite_input` should forced be `False` (or potentially an exception raised).

pyfftw/interfaces/_utils.py
hgomersall: Right, but what does that mean?

pyfftw/interfaces/_utils.py
```diff
@@ -92,29 +102,29 @@ def _Xfftn(a, s, axes, overwrite_input, planner_effort,
...
-        a_copy = a.copy()
+        # If we're going to create a new FFTW object and are not
+        # working with a copy, then we need to copy the input array to
+        # preserve it, otherwise we can't actually  take the transform
+        # of the input array! (in general, we have to assume that the
+        # input array will be destroyed during planning).
+        if not work_with_copy:
+            a_copy = a.copy()
+
+            planner_args = (a, s, axes) + args
```
mforbes: This should not be indented - typo that caused the tests to fail.

pyfftw/interfaces/_utils.py
mforbes: Please let me know what type of exception I should raise.  `ValueError`?

hgomersall: Yup, `ValueError` along with a suitable string. I think it's time to start doing this in a test driven way rather than try to retro fit the tests. I suggest doing the tests first before doing any more.

hgomersall: To be clear, this case should be tested as well.

pyfftw/interfaces/_utils.py
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 #
 # Copyright 2014 Knowledge Economy Developments Ltd
-# 
+#
```
hgomersall: Add yourself to the copyright list.

mforbes:
Okay.  I just wanted to make sure you were on board with the philosophy of this PR.  I will clean things up and create two clean commits - one with failing tests demonstrating the issue, and a second with the resolution.  Should I create a new PR with those, or just force them on top of the current changes for this PR.  (Sorry, I usually use mercurial and am not sure how githug PR's deal with vanishing changesets.)

As for the order: I think I will just leave the numpy default.

hgomersall:
It's up to you re the history. You can clean it up easily with [`git merge --squash`](http://stackoverflow.com/questions/5308816/how-to-use-git-merge-squash).

Other than that, definitely on board with the philosophy. Thanks for your efforts.

mforbes:
@hgomersall It is quite challenging to write a unit test as you suggest because on [line 192](https://github.com/pyFFTW/pyFFTW/blob/master/test/test_pyfftw_numpy_interface.py#L192) a copy is made of the input array which is always `writeable`.  Any suggestions?  (I did not want to go modifying a bunch of the test code since I do not understand it yet.)

hgomersall:
yes good point. It prob needs a bit more fiddling to get to work. I think the solution is to wrap the copies (at least the ones in the actual calls) in the `array_type` lambda. Note that the line you pointed to is the reference (it's should prob not be called test output!).

mforbes:
Also: how do you test for exceptions being raise under appropriate cases (I use `py.tests.raises` or doctests but did not see any examples.)

hgomersall:
like the test in [`test_planner_effort`](https://github.com/pyFFTW/pyFFTW/blob/master/test/test_pyfftw_numpy_interface.py#L499).

mforbes:
I do not understand your proposed solution for the unit test.  I think everything else is ready for review.

test/test_pyfftw_numpy_interface.py
```diff
@@ -1,5 +1,5 @@
 # Copyright 2015 Knowledge Economy Developments Ltd
-# 
+#
```
mforbes: Gah!  Sorry, trailing white space changes... I will go fix.

mforbes:
One final comment: Perhaps the code in `builders._utils` should be updated instead/as well?  It seems to do lots of checks and make copies... one might be able to rely on those instead... I do not understand it deeply enough right now though.

test/test_pyfftw_numpy_interface.py
```diff
@@ -556,6 +556,47 @@ def test_input_maintained(self):
...
+        numpy.random.seed(1)
+        a = numpy.random.random((100,)) + 1j
+        interfaces.numpy_fft.fft(a)
+        a = numpy.random.random((4, 4)) + 1j
+        a.flags.writeable = False
+        assert numpy.allclose(np_fft.fft(a), interfaces.numpy_fft.fft(a))
+
+        self.assertRaisesRegex(
+            ValueError,
+            'Cannot set overwrite_input when not a.flags.writeable',
```
hgomersall: This string would be better IMO as:
'overwrite_input cannot be True when the input array flags.writeable is False'

hgomersall:
`array_type` is just a function that returns an array of the correct type. It can be used on lines 191, 197 and 211 I think.

hgomersall:
`builder._utils` might well benefit from the same treatment. It's a different class of problem though because it is used to do the planning automatically, rather than trying to emulate `np.fft`.

mforbes:
I have to profile, but I suspect that with the current code, there are often spurious copies made by `builder._utils` because the copies made in the numpy interface are not byte-aligned.

hgomersall:
Yes, that is true. Use `pyfftw.zeros_aligned`.

hgomersall:
Is this PR complete now?

mforbes:
@hgomersall I don't understand what your suggestion on how to get the unit tests working.  Can you please give me an example?  Other than that, it should be ready.

hgomersall:
@mforbes I don't quite understand where you're having a problem. Is it related to function that modifies the array?

mforbes:
[Line 192](https://github.com/pyFFTW/pyFFTW/blob/master/test/test_pyfftw_numpy_interface.py#L192) of the test file makes a copy of the array that `array_type` returns, effectively resetting the `writeable` flag.  I do not understand your suggestion for making sure that a locked array gets unit tested.

hgomersall:
Ah yes, it was to wrap the `copy.copy()` inside `array_type`.

Perhaps a better (specifically less risky to legacy code) method is to create a `copy_func` argument to `_validate` that defaults to `copy.copy`, but could be replaced with something like `copy_with_writeable_false` where:

    def custom_copy(array_to_copy):
        array_copy = array_to_copy.copy()
        array_copy.flags.writeable = False
        return array_copy

(defined inside the test method).

Note that `copy.copy()` is used in `_validate` to cope with lists being passed in, which have not `.copy()` method.

hgomersall:
To clarify, then all the instances of `copy.copy()` would then be replaced with `copy_func()`.

hgomersall:
So the signature of `validate` would become:

    def validate(self, array_type, test_shape, dtype, s, kwargs, copy_func=copy.copy):

which is the same as `_validate`.

This would also need to be updated in `test_pyfftw_interfaces_cache.py` which inherits from `InterfacesNumpyFFTTestFFT` in `test_pyfftw_numpy_interface.py`.

mforbes:
Is this what you had in mind?

hgomersall:
Yes, exactly that :) Is this PR now complete?

test/test_pyfftw_numpy_interface.py
```diff
@@ -556,6 +557,51 @@ def test_input_maintained(self):
...
+        a = numpy.random.random((100,)) + 1j
+        interfaces.numpy_fft.fft(a)
+        a = numpy.random.random((4, 4)) + 1j
+        a.flags.writeable = False
+        assert numpy.allclose(np_fft.fft(a), interfaces.numpy_fft.fft(a))
+
+        self.assertRaisesRegex(
+            ValueError,
+            'overwrite_input cannot be True when the ' +
+            'input array flags.writeable is False',
```
hgomersall: Quick FYI, you don't need the '+' here as the string concatenation is implicit on new line. No problem leaving as is though.

test/test_pyfftw_numpy_interface.py
```diff
@@ -556,6 +557,51 @@ def test_input_maintained(self):
...
+                self.validate(array_type,
+                              test_shape, dtype, s, kwargs,
+                              copy_func=copy_with_writeable)
+
+    def test_regression_issue_92(self):
+        '''Quick regression test for issue 92.
+
+        The previous unit tests do not cover this when quick_tests is
+        run because one needs the complex dtype for failure.
+        '''
```
hgomersall: I don't understand this doc string. Could you clarify? This seems to be doing some testing for both the assertion and the `writeable=False` case.

test/test_pyfftw_numpy_interface.py
mforbes: Implicit concatenation was [slated for deprecation](https://www.python.org/dev/peps/pep-3126/), so I trained myself to stop using it, though it will probably be kept until at least python 4.

test/test_pyfftw_numpy_interface.py
mforbes: Ah, sorry.  This was the test I had to implement before `validate` was fixed.  It should be removed and left just for the assertion.

mforbes:
I think it should be ready now.

test/test_pyfftw_numpy_interface.py
hgomersall: Oh, I didn't know that. I noticed the PEP was rejected though.

hgomersall:
Thanks for all your work on this.

Issue #39, opened 2014-03-04
real transform support


drwells:
I am interested in wrapping the real-to-real (at least the discrete sine and discrete cosine) transforms. I set up a branch on my fork of pyFFTW and have written an implementation that appears to work. It still needs tests.

My main concern is this: since `FFTW_FORWARD` is an integer, but the real flags (like `FFTW_ROD00`) are enums, the 'correct' way to handle both with the `self._directions` attribute is a union pointer.

It is worth noting that this contradicts the work proposed in #35.

JoyMonteiro:
Hello, I looked for the code with r2r, but it does not seem to be present anywhere. I looked at drwells'
fork, and could not find them either. Any place I can find it. We are planning to use it in a pseudospectral code, so the dct will get tested quite thoroughly ;)

TIA,
Joy

drwells:
It is available under my `r2r` branch.

Pseudospectral methods were my original motivation to write these bindings (although the pseudospectral code is proprietary, alas).

JoyMonteiro:
Thanks! will check them out.

rabernat:
Sorry for resurrecting this old issue, but I thought it would be better than starting a new, redundant one.

Thanks so much for this awesome package! We have built a psuedo-spectral geostrophic turbulence model which uses pyfftw (version 0.9.2).
https://github.com/pyqg/pyqg

We would like to use discrete sine and cosine transforms (DST, DCT) in order to accommodate different boundary conditions. These transforms are supported by FFTW but not pyfftw.

Will the modifications made by @drwells ever be merged into pyfftw? And am I correct that those modifications include DST and DCT support?

drwells:
I don't mind picking this up again (though I am more interested in the MPI features of FFTW at the moment).

Off topic: I am amused that you are using this for the QGEs (thats what I did too). Here, have a movie I made about a year ago:

http://www.math.vt.edu/people/drwells/qge2.webm

A large amount of the code I used to generate that is proprietary, but not all of it is (I have a pretty sophisticated time integrator). I should take a look into what you have done.

rabernat:
Thanks for your reply! If pyfftw doesn't end up supporting DCT, we will probably end up eliminating the dependency and wrapping FFTW directly from within our project. I would prefer to build on existing work obviously, hence my comment above.

Our goal is to build a general purpose python qg model for both teaching and research purposes. It is well documented and stable now, but still lacking lots of features.
http://pyqg.readthedocs.org/

[...]

PR #95, opened 2016-04-09
R2R try two

drwells:
I spent a few hours rebasing and I am pretty happy with the result. In particular the history is now nice and clean. All of the tests should pass at each commit, so you can review it that way.

I wrote this with the intent of not changing the user-facing API at all (though some details of the `FFTW` class had to change). Therefore this does add some (documented) warts to the library, like those around `_Xfftn`.

### DIFF
--- a/pyfftw/builders/_utils.py
+++ b/pyfftw/builders/_utils.py
@@ -5,8 +5,6 @@
 # 
 # Henry Gomersall
 # heng@kedevelopments.co.uk
-# David Wells
-# drwells <at> vt.edu
 #
 # All rights reserved.
 #
@@ -60,6 +58,8 @@ __all__ = ['_FFTWWrapper', '_rc_dtype_pairs', '_default_dtype', '_Xfftn',
 _valid_efforts = ('FFTW_ESTIMATE', 'FFTW_MEASURE', 
         'FFTW_PATIENT', 'FFTW_EXHAUSTIVE')
 
+_real_to_real_dtypes = [numpy.dtype('float32'), numpy.dtype('float64'),
+                        numpy.dtype('longdouble')]
 # Looking up a dtype in here returns the complex complement of the same
 # precision.
 # It is necessary to use .char as the keys due to MSVC mapping long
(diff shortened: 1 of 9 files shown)


drwells:
The fact that these commits are spread across two years really confuses github...

hgomersall:
This is great work and a really valuable contribution. As it stands though, there are a few things that concern me before merging:

1) The tests are rather sparse. In particular, AFAICT, there are no tests at all for the new builder functions. The tests you have included test the functionality of the new code but very little in the way of testing failure modes and edge cases. I appreciate this is a huge pain to do completely, but the effort on the existing code base has proved to yield a very robust almost bug-free library (I think there is one proper bug in the issue list at the moment). My strategy would have been to re-purpose as much of the existing testing infrastructure as possible (this is how all the other transforms are tested, and I don't think r2r is so radically different) - the classes were designed very much with that in mind. Also, those tests should be manually validated as failing when the functionality is removed.

As an aside, I've actually picked up more than one bug in FFTW itself through the thorough testing.

2) I think we really need to move across to the API discussed in #39 - specifically, deprecate the `direction` parameter and use a `transform_type` argument with suitable enumerations. `direction` is IMO a confusing misnomer when it comes to r2r (and has been for a long time anyway). This would improve `_Xfftn` as well.

3) The interfaces bit should be complete before merging, and as close as possible to `scipy.fftpack`. `scipy.fftpack.dct(a, type=1, norm='ortho')` fails with `NotImplementedError`, but `pyfftw.interfaces.scipy_fftpack.dct(a, type=1, norm='ortho')` will quietly ignore the norm flag. This should also be acknowledged in the test suite.

Similarly, the `n` parameter not `None` raises. Can the existing `_FFTWWrapper` not be modified quite easily? (it just wraps the `FFTW` object with suitable slicers).

4) The documentation needs updating.

As I said, this is valuable work so I sincerely encourage you to persist with this! Thanks!

drwells:
Thank you for the careful review. I agree on all points. It will take me some time to get through this but I will put the time into it (as opposed to the two year gap between now and when #39 was opened).

As far as API changes go: this is your project and I will conform to whatever view you have about the future of the `direction` parameter. I think the best way to handle that would be to add a small, separate change to the library where we make `direction` a deprecated alias of `transform_type` and then I will rebase this branch against that.

Is `_Xfftn` purely internal? If so I will probably change its argument list, unless the aforementioned `transform_type` change beats me to it.

Regardless, here is a TODO list for this branch:
- [ ] Tests for builder functions
- [ ] Finish the `scipy` interface (I foolishly thought I had finished this in 2014 and did not check back)
- [ ] Improve `_Xfftn`
- [ ] Update documentation to describe real to real features

hgomersall:
Yes, I concur with this. The tweak to the API can be made quite easily. It's enough for the time being for you I think to just add the extra parameter (ignoring the deprecation issue). At some point soon I'll make sure everything is consistent and add suitable deprecation warnings and policy for conflicting args as such like.

It's not *my* project as such, I depend on consensus, so feel free to dissent. :)

hgomersall:
Apologies, I just ticked your todo list -  I wasn't aware of that feature! (and I don't know what state it was in because the page hadn't updated).

drwells:
You done broked it!

drwells:
I have not really looked at this in almost a year: I have been working on completely unrelated projects and this has not been a priority for me.

ping @crowsonkb @jo1763 @d-mueller: the three of you are interested in this work. Would you be willing to write the missing tests (e.g., tests for the builder functions and tests for the failure modes)? I believe these are the main blocks from this getting merged (we can raise `NotImplementedError`s for padding and truncation, though it would be nice to have those). Thoughts?

hgomersall:
*poke* still general enthusiasm for merging this work if we can deal with the outstanding blockers.

drwells:
You're right; this has been open too long. I'll wrap this up in January when I can sit down for a few days and get caught up on other changes made to this library.

hgomersall:
Great! There are a few changes that are finding their way in that might influence this. Specifically the work on supporting different sets of precision.

drwells:
I don't see any open PRs or issues on supporting different precision sets; where is this being discussed?

also:
> 
> Conflicting files
> pyfftw/builders/_utils.py
> pyfftw/builders/builders.py
> pyfftw/interfaces/_utils.py
> pyfftw/interfaces/scipy_fftpack.py
> pyfftw/pyfftw.pyx
> sphinx/tutorial.rst
> test/test_pyfftw_scipy_interface.py

*shudders*

hgomersall:
Oops, sorry. Thinks like #189 and #185.

drwells:
Thats right; the shared object files have different names for different precisions. Its been awhile.

hgomersall:
Closing in light of #256 being approved. Thanks @drwells

PR #102, opened 2016-05-01
DOC: use sub-sub-section headings

endolith:
I think this will make them linkable?  So `https://hgomersall.github.io/pyFFTW/sphinx/tutorial.html#wisdom` will work, for instance.

http://docutils.sourceforge.net/docs/user/rst/quickstart.html#sections


hgomersall:
thanks!

PR #104, opened 2016-05-04
Fix LDFLAGS in OS X install instructions

stephentu:
LDFLAGS should point to /usr/local/lib instead of /usr/local/include.


hgomersall:
Yup, thanks!

PR #108, opened 2016-05-17
add option to select orthonormal (unitary) DFT scaling

grlee77:
I am usually interested in applying a unitary DFT operator (i.e. one that satisfies Parseval's theorem).  This differs from the current FFTW default only in the scaling of the forward and inverse transforms:

**ortho (unitary) case:**  scale both forward and inverse by 1/srqt(N)
**current pyFFTW (and Matlab) default:**  forward unscaled,  inverse scaled by 1/N

If the proposed approach seems acceptable, I can also add modifications to the `numpy.fft` interfaces to support the `norm` keyword that was added in numpy 1.10.  In recent numpy, if `norm='ortho'` the behavior matches what I have proposed here.

### DIFF
--- a/pyfftw/pyfftw.pyx
+++ b/pyfftw/pyfftw.pyx
@@ -1359,8 +1359,11 @@ cdef class FFTW:
 
         If ``ortho`` is ``True``, then the output of both forward
         and inverse DFT operations is scaled by 1/sqrt(N), where N is the
-        product of the lengths of input array on which the FFT is taken.  If
-        ortho is True then the 2-norm of fft(A) will equal the 2-norm of A.
+        product of the lengths of input array on which the FFT is taken.  This
+        ensures that the DFT is a unitary operation, meaning that it satisfies
+        Parseval's theorem (the sum of the squared values of the transform
+        output is equal to the sum of the squared values of the input).  In
+        other words, the energy of the signal is preserved.
 
         If either ``normalise_idft`` or ``ortho`` are ``True``, then
         ifft(fft(A)) = A.
--- a/pyfftw/pyfftw.pyx
+++ b/pyfftw/pyfftw.pyx
@@ -654,6 +654,7 @@ cdef class FFTW:
     cdef object _flags_used
 
     cdef double _normalisation_scaling
+    cdef double _sqrt_normalisation_scaling
 
     cdef int _rank
     cdef _fftw_iodim *_dims
@@ -977,6 +978,7 @@ cdef class FFTW:
 
         self._N = total_N
         self._normalisation_scaling = 1/float(self.N)
+        self._sqrt_normalisation_scaling = np.sqrt(self._normalisation_scaling)
 
         # Now we can validate the array shapes
         cdef validator _validator
@@ -1337,8 +1339,9 @@ cdef class FFTW:
             free(self._howmany_dims)
 
     def __call__(self, input_array=None, output_array=None,
-            normalise_idft=True):
-        '''__call__(input_array=None, output_array=None, normalise_idft=True)
+            normalise_idft=True, ortho=False):
+        '''__call__(input_array=None, output_array=None, normalise_idft=True,
+                    ortho=False)
 
         Calling the class instance (optionally) updates the arrays, then
         calls :meth:`~pyfftw.FFTW.execute`, before optionally normalising
@@ -1353,6 +1356,14 @@ cdef class FFTW:
         scaled by 1/N, where N is the product of the lengths of input array on
         which the FFT is taken. If the direction is ``'FFTW_FORWARD'``, this
         flag makes no difference to the output array.
+
+        If ``ortho`` is ``True``, then the output of both forward
+        and inverse DFT operations is scaled by 1/sqrt(N), where N is the
+        product of the lengths of input array on which the FFT is taken.  If
+        ortho is True then the 2-norm of fft(A) will equal the 2-norm of A.
+
+        If either ``normalise_idft`` or ``ortho`` are ``True``, then
+        ifft(fft(A)) = A.
 
         When ``input_array`` is something other than None, then the passed in
         array is coerced to be the same dtype as the input array used when the
@@ -1393,6 +1404,10 @@ cdef class FFTW:
         need the data to persist longer than a subsequent call, you should
         copy the returned array.
         '''
+
+        if ortho and normalise_idft:
+            raise ValueError('Invalid options: ortho and normalise_idft cannot'
+                             ' both be True.')
 
         if input_array is not None or output_array is not None:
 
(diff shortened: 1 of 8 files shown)


grlee77:
to clarify the above:  I have not modified the default behaviour of pyFFTW here, only added an optional keyword argument to choose the `ortho` normalization instead if desired

grlee77:
not sure yet what is causing the Travis failure on Python 3.3 only.  It does not seem obviously related to this PR

hgomersall:
This looks very useful and the right implementation. I'll be happy to merge once the test pass.

It _is_ the new test that is failing: https://travis-ci.org/pyFFTW/pyFFTW/jobs/130986358

I suggest testing in a local 3.3 install. I can't immediately see what the problem is, beyond the warning:
```
/home/travis/build/pyFFTW/pyFFTW/test/test_pyfftw_call.py:401: RuntimeWarning: invalid value encountered in true_divide

  ref_output = ifft(normalise_idft=False).copy()/numpy.float64(ifft.N)

/home/travis/build/pyFFTW/pyFFTW/test/test_pyfftw_call.py:402: RuntimeWarning: invalid value encountered in multiply
```

hgomersall:
To be clear, if this a problem with an unrelated bug, I'm keen to know that explicitly before merging.

grlee77:
The failing test is `test_call_with_normalisation_precision` which is **not** a new test, but it is in the same file and immediately above where I added my new tests.

I created a test environment via
` conda create -n py33 python=3.3 numpy scipy cython setuptools`
and numpy et al seem to be the exact same version as on Travis, but I am unable to reproduce the test failure.  I am on ubuntu 14.04.

grlee77:
Probably unreleated to this bug, but looking more closely at `test_call_with_normalisation_precision` it seems like `self.fft()` should be called prior to the calls to `ifft` so that the comparisons later in the test aren't being done on arrays that are all zeros.

hgomersall:
Oops, you're quite right. Do you fancy fixing the test bug you noted? It seems to be an inconsistent failure - I'm not sure why exactly.

grlee77:
All green now!
Updating the numpy interfaces was pretty easy.  It was actually more difficult updating the tests!

I never really determined what caused the original Travis failure way up above, but it seems to have been a fluke.  From the error message it sounds like ifft.N was zero somehow, but that shouldn't be possible.

I can rebase and squash a few of the minor/testing commits here once this is deemed ready to go.

hgomersall:
Ok, #109 is merged. No need to rebase if you'd rather just pull in upstream, it's up to you.

Is this complete now as far as you're concerned?

pyfftw/pyfftw.pyx
```diff
@@ -1354,6 +1357,14 @@ cdef class FFTW:
         which the FFT is taken. If the direction is ``'FFTW_FORWARD'``, this
         flag makes no difference to the output array.
 
+        If ``ortho`` is ``True``, then the output of both forward
+        and inverse DFT operations is scaled by 1/sqrt(N), where N is the
+        product of the lengths of input array on which the FFT is taken.  If
+        ortho is True then the 2-norm of fft(A) will equal the 2-norm of A.
```
hgomersall: A little extra info would be helpful here. Something like: `i.e. energy is preserved across the transform`.

hgomersall: Not sure what the best terminology is, but 2-norm is a bit opaque to many people so some clarification is helpful I feel.

pyfftw/pyfftw.pyx
```diff
@@ -1394,6 +1405,10 @@ cdef class FFTW:
         copy the returned array.
         '''
 
+        if ortho and normalise_idft:
+            raise ValueError('Invalid options: ortho and normalise_idft cannot'
+                             ' both be True.')
+
```
hgomersall: hmmm I'm not super keen on this - it's a bit mucky to have a requirement for exclusive arguments. However, I can't see how to do it cleanly whilst maintaining backwards compatibility.

hgomersall:
It looks basically good to me. Thanks for your work on it, including the necessary documentation additions. There are a couple of comments in the code.

pyfftw/pyfftw.pyx
grlee77: Agreed.  The alternative I considered was a norm kwarg that could be one of `{'ortho', 'idft', None}`, but that is not backwards compatible.

grlee77:
@hgomersall 
Please review the requested doc string changes in the final commit here.  Hopefully it is clearer now.

I rebased on current master and squashed two of the minor commits where I was debugging.  Nothing was changed or added aside from the most recent DOC commit.

pyfftw/pyfftw.pyx
hgomersall: An alternative is to do `normalise_idft='ortho'` or `normalise_idft=True` or `normalise_idft=None`. Is that more mucky?

grlee77: The `ortho` scaling is applied both to the dft and idft which makes the `normalise_idft` variable name slightly misleading.  I don't have too strong of an opinion either way, though.  Day to day I am usually using either the builders or the numpy interfaces so I don't call FFTW directly.

hgomersall: Yeah, I concur. I think I'll accept as is for the time being and review a bit later. We need an API consolidation at some point - there are a few things I've been mulling over that need changing.

hgomersall: Further to your last point - presumably you _do_ use the FFTW `__call__` interface if you're using builders - they just return an FFTW object.

hgomersall:
Merged, thanks!

jakirkham:
Did this get added to the docs?

hgomersall:
Well, it's in HEAD.

jakirkham:
The docs are? Or this change? As this change seems to have been present since 0.10.2.

hgomersall:
What? The docs are in head, the code I think is a bit older (see the merge requests above...)

grlee77:
There is only indirect documentation for the interfaces which link to the `numpy` documentation where the `norm` parameter is documented (although the doc hosting at `scipy` is having issues at the moment):
http://pyfftw.readthedocs.io/en/latest/pyfftw/interfaces/numpy_fft.html

Otherwise there is the docstring text in the `__call__` method here:
https://github.com/pyFFTW/pyFFTW/blob/530e400e7c45a8111bf9aec170d10cacb4d0d051/pyfftw/pyfftw.pyx#L1354-L1366

Is there another location you would like to see this behavior documented?

jakirkham:
I didn't actually know that there was documentation on RTD. Had only seen the docs on GitHub pages, which seem a bit out-of-date. Should there be a link to RTD in the Readme (or did I simply miss it)?

On a different point, I'm not sure if the RTD are working. When I click on the link above I see a title heading, but no API info. 😕

hgomersall:
No, AFAICT, RTD is broken in conjunction with my code. The problem was I couldn't get RTD to auto generate from Cython code (if I recall correctly). Github pages should be more-or-less up to date with the release. It's still an unacceptably big hassle for me to make a release. I put some time in to try and fix this a while ago but didn't quite finish.

grlee77:
Okay, I was mistaken about the API docs.  The API docs at Github pages are rendered properly, but are just a bit outdated (currently they appear to be for release `0.10.1`).  I think if they were updated to `0.10.4` than the `norm` parameter would be shown there as well.

PR #109, opened 2016-05-18
fix two minor test bugs

grlee77:
I came across these while working on #108.  I thought it would be cleaner to make a separate PR for these.  I can always rebase #108 after this is merged.


hgomersall:
Well that might well explain the test failure - if `fft` is never called, the output array will just contain random junk with whatever happens next being not so well defined. Thanks for this fix.

Issue #106, opened 2016-05-10
Builders and axes


davidbrough1:
The following code runs with pyfftw 0.9.2 and 0.10.1 without a problem

    import numpy as np
    from pyfftw import builders

    X = np.ones((1, 33, 33, 2))

    FX = builders.rfftn(X, axes=[1, 2, 3])

but if the `axes` argument is changed 

    import numpy as np
    from pyfftw import builders
 
    X = np.ones((1, 33, 33, 2))

    FX = builders.rfftn(X, axes=[1, 2])

a  `RuntimeError` is raised

    Traceback (most recent call last):
      File "pyfftw_issues.py", line 6, in <module>
        FX = builders.rfftn(X, axes=[1, 2])
      File "/home/david/anaconda2/envs/pymks_dev_fftw/lib/python2.7/site-packages/pyfftw/builders/builders.py", line 471, in rfftn
        avoid_copy, inverse, real)
      File "/home/david/anaconda2/envs/pymks_dev_fftw/lib/python2.7/site-packages/pyfftw/builders/_utils.py", line 169, in _Xfftn
        flags, threads)
      File "pyfftw.pyx", line 971, in pyfftw.pyfftw.FFTW.__cinit__ (pyfftw/pyfftw.c:7658)
    RuntimeError: The data has an uncaught error that led to the planner returning NULL. This is a bug.

grlee77:
Do you by any chance also have MKL installed? (e.g. using anaconda with the default numpy/scipy)  If so, this is likely another case of #40

(For your particular example above, the MKL FFT routines don't support transforms where the number of axes to be transformed are more than 1 fewer than the number of array dimensions)

hgomersall:
Yeah, it looks like that - I certainly can't reproduce it. I suggest following the workarounds in #40 and if that's the case, go and bug Anaconda about it.

davidbrough1:
It looks like it might be a case similar to #40.

For what it's worth, when using a version of anaconda <=2.4.1 with MKL I don't get this error regardless of the order of import statements. If I use a version of anaconda >2.4.1 with MKL I have to have the import statements in the following order to not get the error.

     from pyfftw import builders
     import numpy as np

ferdinandvanwyk:
Just started seeing this error on Travis CI since I use Miniconda to handle packages which may require compilation. So others don't have to follow a trail of issues, in addition to the above solutions you can simply disable MKL optimizations if you don't need them in Anaconda/Miniconda via:

[...]

Issue #40, opened 2014-04-21
RuntimeError for N-D arrays on anaconda distribution with mkl optimizations


jonaballe:
Hello.

I have started to rely heavily on the anaconda python distribution because of its simple maintenance and because of their MKL optimized linear algebra code in numpy/scipy (which is free for academic use). In terms of FFT implementations, however, they provide nothing compared to FFTW, which is why I use pyFFTW for that (and I want to say a big thank you for providing this well-designed set of wrappers!).

Unfortunately, I just ran into the following error (in pyfftw.pyfftw.FFTW.**cinit** (pyFFTW/pyfftw/pyfftw.c:7613):

"RuntimeError: The data has an uncaught error that led to the planner returning NULL. This is a bug."

It is reproducible with the following three lines of code:

from numpy import empty
from pyfftw.interfaces.numpy_fft import rfft2
rfft2( empty( ( 2, 2, 50, 50 ) ) )

The exception is thrown only when I'm using the MKL optimization package. Everything is fine with arrays of up to three dimensions.

I am somewhat puzzled as to why this happens, and not sure whether it is a bug in the anaconda distribution or in pyfftw (or fftw). But having fast linear algebra and fast fft at the same time is critical for my work. As I am unaware of the internal structure of pyfftw, can you assist me in figuring out this problem?

Thanks,
Johannes

hgomersall:
It's hard to see what we can do from the pyfftw end for this. If the MKL is exporting symbols that another library uses (specifically to be API compatible), then it's sort of the MKL's job to not cause problems. I'm inclined to mark this `wontfix` and document the workaround.

hgomersall:
Actually, is this still a problem? I think it's probably a job for anaconda if it is.

jonycgn:
so far, i've been using the -Bsymbolic workaround successfully. documenting the problem with the workaround seems appropriate ... so as for me, you can go ahead and close the issue.

hgomersall:
@drwells You seem to suggest that Anaconda just works when the import is in the right order. Is this correct? When is the -Bsymbolic workaround necessary?

drwells:
That was nearly two years ago; I don't remember. I'll try and reproduce this in the next week.

jonycgn:
the workaround should be necessary in general, since you don't really know what other modules that you are importing depend on, and in what order they import stuff.

aganders3:
[...]

PR #116, opened 2016-06-03
BLD: use optional STATIC_FFTW_DIR environment variable to link static…

grlee77:
I am trying to get a version of conda packages for `fftw` and `pyFFTW` that do suffer from the MKL symbol clashing issues on linux that were reported in #40 and #106.  I am taking the approach proposed by @jonycgn where the FFTW libs are compiled with the -fPIC flag so that they can be statically linked.

The current `fftw` package in my PR at conda-forge/fftw-feedstock#4 builds both static and dynamic libraries.  A problem I encountered is that the current pyFFTW `setup.py` will link to the dynamic libraries rather than the desired static ones so that the suggested solution of `CFLAGS="-I${PREFIX}/include -Xlinker -Bsymbolic" pip install pyFFTW` does not work.

This PR modifies `setup.py` to use the absolute path to the static FFTW libs if a STATIC_FFTW_DIR environment variable is defined.  I would appreciate if you have other ideas of a better way to do this.

If the following two exports are made prior to installing pyFFTW, the problems with MKL should not occur.
1.) export STATIC_FFTW_DIR=${PREFIX}/lib  where ${PREFIX} is the base of the current anaconda environment with the `fftw` package installed.
2.) export CFLAGS="$CFLAGS -Wl,-Bsymbolic"

A `pyFFTW` package that is built in this manner would then be made available through [condo-forge](https://conda-forge.github.io).  Rather than use a custom patch to setup.py in the build recipe, I thought I would try a PR here first.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -142,7 +142,7 @@ def get_extensions():
         common_extension_args['libraries'] = []
     else:
         # otherwise we use dynamic libraries
-        extra_link_args = []
+        common_extension_args['extra_link_args'] = []
         common_extension_args['libraries'] = libraries
 
     ext_modules = [
--- a/setup.py
+++ b/setup.py
@@ -53,6 +53,9 @@ MICRO = 1
 ISRELEASED = False
 VERSION = '%d.%d.%d' % (MAJOR, MINOR, MICRO)
 
+static_fftw_path = os.environ.get('STATIC_FFTW_DIR', None)
+link_static_fftw = static_fftw_path is not None
+
 def get_package_data():
     from pkg_resources import get_build_platform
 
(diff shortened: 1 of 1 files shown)


hgomersall:
I think this looks fine and I'm happy to merge the logic as is. However, could you tweak the code slightly to make it easier to parse. So `get_extensions` would be something like:

```python
def get_extensions():
    from distutils.extension import Extension

    common_extension_args = {
        'include_dirs': get_include_dirs(),
        'library_dirs': get_library_dirs()}

    try:
        from Cython.Build import cythonize
        sources = [os.path.join(os.getcwd(), 'pyfftw', 'pyfftw.pyx')]
        have_cython = True

    except ImportError as e:
        # no cython
        sources = [os.path.join(os.getcwd(), 'pyfftw', 'pyfftw.c')]
        if not os.path.exists(sources[0]):
            raise ImportError(
                str(e) + '. ' +
                'Cython is required to build the initial .c file.')

        have_cython = False
    
    libraries = get_libraries()
    if link_static_fftw:
        from pkg_resources import get_build_platform
        if get_build_platform() in ('win32', 'win-amd64'):
            lib_pre = ''
            lib_ext = '.lib'
        else:
            lib_pre = 'lib'
            lib_ext = '.a'
        extra_link_args = []
        for lib in libraries:
            extra_link_args.append(
                os.path.join(static_fftw_path, lib_pre + lib + lib_ext))
        # now that full paths to libraries are in extra_link_args remove them
        # from common_extension_args
        common_extension_args['extra_link_args'] = extra_link_args
        common_extension_args['libraries'] = []
    else:
        common_extension_args['extra_link_args'] = []
        common_extension_args['libraries'] = libraries

    ext_modules = [
        Extension('pyfftw.pyfftw', sources=sources,
                         **common_extension_args)]
<snip>
```

hgomersall:
That should be equivalent. If you are happy with those changes and implement them, I'll merge immediately.

grlee77:
Thanks for the quick review.
I went ahead and moved the STATIC_FFTW_DIR environment variable check into that function as well

grlee77:
I can squash the commits if you prefer

hgomersall:
No need. Thanks.

PR #126, opened 2016-06-06
Fix for all the recent version number breakages. It'll be necessary t…

hgomersall:
…o revisit this at some point.


anntzer:
FWIW you may want to have a look at https://github.com/warner/python-versioneer.

hgomersall:
@anntzer  Yes, that looks very useful. I hadn't come across it before (I was going to play around with a tool that did things slightly differently).

Thanks!

PR #140, opened 2016-12-29
ENH: process float16 inputs in single precision

grlee77:
This PR adds a special case for  `numpy.float16` data, causing it to be processed in single rather than double precision.  I think it makes sense to use the nearest available floating point precision in FFTW.  

a simple example:

```python
import numpy as np
from pyfftw.interfaces.numpy_fft import fftn
a = fftn(np.random.randn(32).astype(np.float16))
print(a.dtype)
```
`a` in the example above is `complex64` with this PR vs  `complex128` in current master.

If this seems acceptable, can you please point to where the best place to add a test for this would be?

### DIFF
--- a/pyfftw/builders/_utils.py
+++ b/pyfftw/builders/_utils.py
@@ -99,12 +99,21 @@ def _Xfftn(a, s, axes, overwrite_input,
 
     # Make the input dtype correct
     if a.dtype.char not in _rc_dtype_pairs:
-        # We make it the default dtype
-        if not real or inverse:
-            # It's going to be complex
-            a = numpy.asarray(a, dtype=_rc_dtype_pairs[_default_dtype.char])
+        if a.dtype == numpy.dtype('float16'):
+            # convert half-precision to single precision rather than double
+            if not real or inverse:
+                a = numpy.asarray(
+                    a, dtype=_rc_dtype_pairs[numpy.dtype('float32').char])
+            else:
+                a = numpy.asarray(a, dtype=_default_dtype.char)
         else:
-            a = numpy.asarray(a, dtype=_default_dtype)
+            # We make it the default dtype
+            if not real or inverse:
+                # It's going to be complex
+                a = numpy.asarray(
+                    a, dtype=_rc_dtype_pairs[_default_dtype.char])
+            else:
+                a = numpy.asarray(a, dtype=_default_dtype)
 
     elif not (real and not inverse) and not a_is_complex:
         # We need to make it a complex dtype
--- a/pyfftw/builders/_utils.py
+++ b/pyfftw/builders/_utils.py
@@ -105,7 +105,7 @@ def _Xfftn(a, s, axes, overwrite_input,
                 a = numpy.asarray(
                     a, dtype=_rc_dtype_pairs[numpy.dtype('float32').char])
             else:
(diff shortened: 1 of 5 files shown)


hgomersall:
As it stands, this adds an untested path. To merge we need to add this to the spec/test-suite.

It should be quite easy to do...

grlee77:
sure.  just wanted to see if you were open to the idea before spending time making tests

grlee77:
I had not run the scipy tests prior to uploading and didn't realize they import from the numpy ones as for example:

real_dtypes = test_pyfftw_numpy_interface.real_dtypes

If real_dtype is float16 this causes scipy's fftpack to raise an error.  will see about a workaround for this.

grlee77:
We cannot validate vs. scipy.fftpack for float16 input as scipy will raise an error.  e.g.

```python
import numpy as np
import scipy.fftpack

scipy.fftpack.fft(np.ones(8).astype(np.float16))

  File "<ipython-input-5-3b66b4d7b5f7>", line 1, in <module>
    scipy.fftpack.fft(np.ones(8).astype(np.float16))

  File "/Users/lee8rx/anaconda/lib/python3.5/site-packages/scipy/fftpack/basic.py", line 257, in fft
    raise ValueError("type %s is not supported" % tmp.dtype)

ValueError: type float16 is not supported
```

grlee77:
I opened a related PR at `scipy` to convert float16->float32 prior to calling the transform: scipy/scipy#6909.

hgomersall:
Where are we with this?

grlee77:
I added some tests for the transforms with float16 input and think it is ready for review.

scipy/scipy#6909 was merged (this lets `scipy.fftpack` convert float16 inputs to single precision rather than raising an error).  Those changes will appear in the pending 0.19 release.

In light of that it may be worth modifying the last commit above (1d49e23), to only override `test_pyfftw_numpy_interface.complex_dtypes` in the case that `scipy.__version__ < 0.19`

hgomersall:
Ok thanks. Please do add the change. I'll make the review when I have a moment!

hgomersall:
Many thanks for this!

hgomersall:
Apparently this broke appveyor. It seems to not be able to find the lib file. I suspect this actually has nothing to do this this commit but is more down to the time between commits.

hgomersall:
FYI https://ci.appveyor.com/project/hgomersall/pyfftw/build/1.0.229/job/3yra8j7dxgb6shon

grlee77:
Yeah this seems unrelated.  The failures appear to be in downloading/finding the FFTW .zip file:
e.g.
```
ERROR: The system cannot find the file specified.
fftw-3.3.4-dll32.zip
```

so likely related to something changing on Appveyor's end?

hgomersall:
Yes seemingly. I've raised a query on the [appveyor support forum](http://help.appveyor.com/discussions/problems/6558-downloading-from-ftp-server-in-cmd-file).

grlee77:
I don't see what is wrong in the current Appveyor script, but it looks like it is also possible to use `curl` to download the file as described here:
https://www.appveyor.com/docs/how-to/download-file/

might be worth a shot as a workaround

hgomersall:
PR in #172.

hgomersall:
(FYI, your suggestion fixed that problem and exposed a new one)

Issue #151, opened 2017-03-24
Dask wrappers/interface for pyFFTW


jakirkham:
Often I use FFTs on large data and am eager to get any speed up I can with it. pyFFTW has been great at delivering really nice performance while also making it very easy to use FFTW thanks to the NumPy style interface.

Though one thing that would be really great is to make it easier still to leverage pyFFTW on data much larger than memory and handle this in parallel. One library that is really fantastic for this is Dask. By allowing one to treat massive files like NumPy arrays and leverage delayed computation, one is able to construct different workflows with large data while keeping focusing on the big picture.

In an attempt to get these two great libraries to play together, @mrocklin, @jcrist, and I worked together on abstracting an FFT interface in Dask ( https://github.com/dask/dask/pull/2093 ) to allow one to easily wrap FFT functions from external libraries that match the NumPy API. While doing this I made sure to test against pyFFTW explicitly. It is trivial for any user to roll these themselves. However, I wondered if it would make sense to included some prewrapped copies of these functions in pyFFTW as part of a new Dask interface for instance.

Would be curious to know your thoughts on this proposal.

hgomersall:
I'm not fully sure of what you're proposing here. Would you suggest an addition to the API, something like `pyfftw.dask`? (also, Dask looks really cool and like its something I should be using!).

jakirkham:
Sorry, was thinking something like `pyfftw.interfaces.dask_numpy_fft` (though we can play with the naming). It would be similar to `pyfftw.interfaces.numpy_fft`, but with Dask support enabled.

hgomersall:
I'm not averse to the idea. Can you enumerate the changes that would be needed?

jakirkham:
Sure. Here are the steps. 

1. Create the `pyfftw.interfaces.dask_numpy_fft` module.
2. Wrap `pyfftw.interfaces.numpy_fft` functions (e.g. `fft = dask.array.fft.fft_wrap(pyfftw.interfaces.numpy_fft.fft)`).
3. Add some tests to exercise these wrappers.

Am more than happy to open a PR with these changes.

hgomersall:
Yeah, do it. Thanks!

jakirkham:
Have added a proposal for adding this Dask interface in PR ( https://github.com/pyFFTW/pyFFTW/pull/154 ).

jakirkham:
[...]

PR #154, opened 2017-03-30
Provide a Dask interface to pyFFTW's 1-D FFTs

jakirkham:
Fixes https://github.com/pyFFTW/pyFFTW/issues/151

Wrap the existing NumPy interface so that it will integrate nicely with Dask. As Dask currently only supports 1-D FFT functions, simply wrap and provide those for now. In the future Dask will support 2-D and N-D FFT function. This will be updated at that time to provide those too. Also add borrow tests from the NumPy interface to use with the Dask interface. Update it to use the Dask interface, work with Dask Arrays, and drop any tests that deal with mutation or 2-D/N-D FFT functions. Further autdoc all functions in the interface so as to include them in the documentation.

Note: The current release of Dask only supports 1-D FFT functions. However, 2-D and N-D FFT functions are supported in `master` and should be available in the next release. We can update this interface as soon as they are available.


jakirkham:
So it seems some really old copies of Dask are used on Python 3.3 and 3.4. Maybe we can just skip these tests there? Python 2.7 passes. Python 3.5 passes the tests we have added, but fails some other seemingly unrelated tests.

pyfftw/interfaces/dask_fft.py
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env python
+#
+# Henry Gomersall
+# heng@kedevelopments.co.uk
```
hgomersall: Perhaps you need to add yourself here...

test/test_pyfftw_dask_interface.py
```diff
@@ -0,0 +1,586 @@
+# Copyright 2015 Knowledge Economy Developments Ltd
+#
+# Henry Gomersall
+# heng@kedevelopments.co.uk
```
hgomersall: ditto

hgomersall:
It looks sensible to me. Is the test file just a modified clone of the np interface test file?

I would like to wrap the Dask stuff with an import exception handler - Dask shouldn't be a global dependency.

This is done  for scipy in the [interfaces `__init__` function](https://github.com/pyFFTW/pyFFTW/blob/master/pyfftw/interfaces/__init__.py#L220)

IMO skipping 3.3 and 3.4 for Dask only is fine. You can use the `skipIf` decorator like [here](https://github.com/pyFFTW/pyFFTW/blob/master/test/test_pyfftw_scipy_interface.py#L100). Adding a comment to explain would be nice too.

hgomersall:
Oh sorry, you made a note about the tests in your first comment.

pyfftw/interfaces/dask_fft.py
jakirkham: Maybe we should just have an author's file instead.

jakirkham:
> It looks sensible to me. Is the test file just a modified clone of the np interface test file?

Yep. Had to cut the 2-D and N-D stuff as it is not supported in a release yet. Will add back 2-D and N-D stuff once a newer release supports it. Also Dask arrays are immutable (they are actually graphs of operations that are lazy evaluated) so any mutable tests had to go as well. Then, of course, some tweaks to use Dask arrays and not NumPy arrays. Though going back and forth is trivial.

> I would like to wrap the Dask stuff with an import exception handler - Dask shouldn't be a global dependency.

Sounds fine to me. Will look into it.

> IMO skipping 3.3 and 3.4 for Dask only is fine. You can use the skipIf decorator like here. Adding a comment to explain would be nice too.

Ok, also we might be able to pip install Dask to get a new enough version that works on Python 3.4 if we want. However they don't test against Python 3.3 any more, so we should definitely skip that one.

jakirkham:
Alright have fixed this up so that Dask is an optional requirement. The interface is hidden if Dask isn't available. Now tests are skipped if the Dask isn't a new enough version, which handles the Python 3.3 and 3.4 issues seen before in a cleaner way. The tests seem to pass.

However, am still seeing a couple seemingly unrelated test failures with Python 3.3 and 3.5 Not sure what is going on there. Any advice definitely appreciated.

jakirkham:
@mrocklin, am working on an interface for Dask to add to pyFFTW. Would generally appreciate any feedback you have on it.

Also am noticing some tests failing with `threading.active_count()` not equal to `1`. Not sure if this would even be related to Dask (or Distributed), but figured I'd reach out in case it is.

hgomersall:
The failure seems not so serious. It's a failure in a pre-test tidy up routine. The cache is running in a different thread so the two failing tests first of all do a pause to let the cache be emptied properly.

I don't know why your changes are suddenly causing it to fail. I've seen this before and solved it by increasing the time allowed to empty the cache (on lines 171 and 196 of [the cache tests](https://github.com/pyFFTW/pyFFTW/blob/master/test/test_pyfftw_interfaces_cache.py).

That said, it seems a bit suboptimal to keep increasing the test times. I suspect there is something to do with CPU contention on the Travis server, but I don't really know. It could very easily be in that dark and murky domain of multi threading interacting with CPU availability to produce non obvious outcomes. I suspect it's not a bug.

I'd be interested to know if the tests run on your local machine...

pyfftw/interfaces/dask_fft.py
hgomersall: Yup, that could work too.

jakirkham:
So if I run the whole test suite, I get 3 failures from `test/test_pyfftw_interfaces_cache.py`. They are all `AssertionError`s. The failing tests and lines are...

* `CacheTest.test_cache_parent_thread_ended`: line 173
* `CacheTest.test_delete_cache_object`: line 197
* `CacheTest.test_objects_removed_after_keepalive`: line 302

hgomersall:
hmmm interesting. I'm slightly at a loss. Can you try disabling and reenabling the cache (instead of the pause)?

jakirkham:
Me too. 😕 Sorry, not super familiar with the cache, how do I do that?

hgomersall:
The pertinent docs are [here](https://hgomersall.github.io/pyFFTW/pyfftw/interfaces/interfaces.html#module-pyfftw.interfaces.cache) and the code is [here](https://github.com/pyFFTW/pyFFTW/blob/master/pyfftw/interfaces/cache.py).

Thanks for your work on this. I appreciate all your contributions.

jakirkham:
Yeah, I gave enabling/disable the cache a try. No dice. 😞

It's interesting as the Dask tests run before the cache tests do. I wonder if one is effecting the other somehow. 😕 Maybe I'll give switching the order a try.

hgomersall:
Yes, I think that probably is happening. If you disable the Dask tests, do the numpy_interface tests run before or after? That would explain why it only appears with the code.

jakirkham:
So I've been using `pytest` to run the test suite piecewise. If I just run the cache interface tests alone, I have no issues. 

If I rename the cache tests to `test_pyfftw_cache_interface.py` (matching the naming of the other interfaces), it is the second test module to run and it only fails with the `test_objects_removed_after_keepalive`, which solved by bumping the timeout (I'm on Mac) and didn't occur on Travis CI before. Note that it follows `test_pyfftw_builders.py`. I don't know if that would cause an issue or not.

hgomersall:
Ok, I think the problem is due to the ordering, with either the dask or the numpy tests causing the cache to be in an unexpected state.

There is only one reference to the cache in any test other than the cache tests, and it's in the numpy interfaces (and now the Dask interfaces) on line 173. I'm minded to try removing that disabling and seeing if it fixes the problem...

hgomersall:
What we need is some debugging info on where that thread is from. As a starting point, we can call `threading.enumerate()` to find out what is running, and then take a peek at those threads that are returned.

hgomersall:
The enumerate of the last comment should be done where the tests are failing.

jakirkham:
I think Dask and/or Distributed is starting daemon threads. So they won't shutdown until the process ends. TBH I don't know if we are going to find a better solution than this workaround. Unless, of course, we run the Dask tests in a totally separate Python process.

mrocklin:
Dask is not shy about starting daemon threads

mrocklin:
My apologies if this has caused some development pain

jakirkham:
Is there another way to shutdown daemon threads that you know of, @mrocklin?

mrocklin:
Shutting down threads generally, no.

I'm curious what threads are causing grief.  Do you have any information
about them?

jakirkham:
It's not that the threads are causing grief per se. Just that they are daemon threads and thus are still around.

The [caching tests]( https://github.com/pyFFTW/pyFFTW/blob/d883f22426aebfc4f6094350d90087a07082a1fc/test/test_pyfftw_interfaces_cache.py ) in pyFFTW have a bunch of assertions checking that an exact number of threads are in use at different times. The asserted values are all going to be wrong if these cache tests are run after Dask as they don't factor in the daemon threads may already be present.

IMHO there are three possible general fixes.

1. Run the cache tests before the Dask tests.
2. Compute the relative number of threads used in the cache tests.
3. Drop the thread count checks.

Right now this PR has included solution 1. If these tests are really flaky (I'm not too familiar with their history), 3 might be reasonable. TBH 2 feels like a lot of work that is a bit out of scope.

hgomersall:
This is interesting problem. I think the better strategy is to change the checks in the failing tests to make they only pick up the ones created by pyfftw. I don't know how to do this, but I suspect there is a reasonable way.

hgomersall:
Something like the following should detect only the pertinent thread I think:

    cache_thread_running = False
    for each_thread in threading.enumerate()
        cache_thread_running = (
            cache_thread_running or each_thread._Thread__target == pyfftw.interfaces.cache._Cache._run)

    assert not cache_thread_running

jakirkham:
TBH I know very little about how the cache works in pyFFTW nor the test suite associated. So I don't feel very comfortable rewriting the whole test suite. While there may be better solutions than the one included here, I don't think I'm the best person to ask to rework the cache's test suite.

hgomersall:
Ok, I can revisit this. I'll try to take a look on Sunday. In essence, what needs to be done is the check on the threads in those two test needs to check only those spawned by pyfftw. I'll make a little internal function that does as described above.

Can you allow me rights to edit your branch? ([like this](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/).

jakirkham:
Thanks @hgomersall. Definitely appreciate it.

Sure, you should have permissions. Please let me know if you have any issues.

hgomersall:
Ok, I've pushed changes that should fix the problem. I'd be interested to know if it does (I think the problem is gone because of the name change, right?).

hgomersall:
That didn't work. Trying again!

jakirkham:
Thanks very much @hgomersall. If I revert the directory rename locally, the issue appears to go away. Will push that change here so we can see how it does on CIs.

jakirkham:
Appears to have worked.

hgomersall:
Good news. Are we happy with this now?

jakirkham:
Yep, LGTM. Thanks again.

hgomersall:
Great, thank you!

pyfftw/interfaces/__init__.py
```diff
@@ -128,6 +128,16 @@
...
 
+:mod:`~pyfftw.interfaces.numpy_fft`
+"""""""""""""""""""""""""""""""""""
+
+* :func:`pyfftw.interfaces.numpy_fft.fft`
+* :func:`pyfftw.interfaces.numpy_fft.ifft`
+* :func:`pyfftw.interfaces.numpy_fft.rfft`
+* :func:`pyfftw.interfaces.numpy_fft.irfft`
+* :func:`pyfftw.interfaces.numpy_fft.hfft`
+* :func:`pyfftw.interfaces.numpy_fft.ihfft`
```
jakirkham: This should have been `dask_fft` instead of `numpy_fft`. PR ( https://github.com/pyFFTW/pyFFTW/pull/166 ) fixes this.

pyfftw/interfaces/dask_fft.py
```diff
@@ -0,0 +1,66 @@
...
+point and so will the returned array. Half precision input will be
+converted to single precision.  Otherwise, if any type conversion is
+required, the default will be double precision.
+
+The exceptions raised by each of these functions are mostly as per their
+equivalents in :mod:`dask.fft`, though there are some corner cases in
+which this may not be true.
+'''
+
+from . import numpy_fft
```
jakirkham: It's probably cleaner to import functions from this interface, wrap them, and overwrite the associate variables with our wrappers. This was we don't accidentally expose this module in our interface. PR ( https://github.com/pyFFTW/pyFFTW/pull/165 ) does exactly this.

Issue #146, opened 2017-02-28
README on pypi still points to github.com/hgomersall/pyfftw


anntzer:
Last line of https://pypi.python.org/pypi/pyFFTW ("The documentation can be found here, and the source is on github.") has an outdated link (compare with the README in the repo).

jakirkham:
Care to do a PR? Would just be changing this [line]( https://github.com/pyFFTW/pyFFTW/blob/530e400e7c45a8111bf9aec170d10cacb4d0d051/setup.py#L460 )? Though the update won't propagate to PyPI until the next release.

jakirkham:
Also these [lines]( https://github.com/pyFFTW/pyFFTW/blob/530e400e7c45a8111bf9aec170d10cacb4d0d051/setup.py#L219-L220 ) probably should be updated too.

anntzer:
Feel free to pick the issue up.

jakirkham:
Please see PR ( https://github.com/pyFFTW/pyFFTW/pull/156 ).

PR #156, opened 2017-03-30
Use org URLs

jakirkham:
Fixes https://github.com/pyFFTW/pyFFTW/issues/146

Tries to update URLs that can be updated to point to the org. Went through and checked to make sure they are correct, but wouldn't mind a second check from someone else.

ATM it seems there is no AppVeyor project for the org. So that URL was not changed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -216,7 +216,7 @@ This should probably work for OSX, though I've not tried it.
 Numpy is a dependency for both.
 
 The documentation can be found
-`here <http://hgomersall.github.com/pyFFTW/>`_, and the source
+`here <http://pyFFTW.github.io/pyFFTW/>`_, and the source
 is on `github <https://github.com/pyFFTW/pyFFTW>`_.
 '''
 
@@ -460,7 +460,7 @@ def setup_package():
         'description': (
             'A pythonic wrapper around FFTW, the FFT library, presenting a '
             'unified interface for all the supported transforms.'),
-        'url': 'http://hgomersall.github.com/pyFFTW/',
+        'url': 'http://pyFFTW.github.io/pyFFTW/',
         'long_description': long_description,
         'classifiers': [
             'Programming Language :: Python',
--- a/setup.py
+++ b/setup.py
@@ -216,7 +216,7 @@ This should probably work for OSX, though I've not tried it.
 Numpy is a dependency for both.
 
 The documentation can be found
-`here <http://pyFFTW.github.io/pyFFTW/>`_, and the source
+`here <http://hgomersall.github.com/pyFFTW/>`_, and the source
 is on `github <https://github.com/pyFFTW/pyFFTW>`_.
 '''
 
@@ -460,7 +460,7 @@ def setup_package():
         'description': (
             'A pythonic wrapper around FFTW, the FFT library, presenting a '
             'unified interface for all the supported transforms.'),
-        'url': 'http://pyFFTW.github.io/pyFFTW/',
+        'url': 'http://hgomersall.github.com/pyFFTW/',
         'long_description': long_description,
         'classifiers': [
             'Programming Language :: Python',
--- a/setup.py
+++ b/setup.py
@@ -216,8 +216,8 @@ This should probably work for OSX, though I've not tried it.
 Numpy is a dependency for both.
 
 The documentation can be found
-`here <http://hgomersall.github.com/pyFFTW/>`_, and the source
-is on `github <https://github.com/hgomersall/pyFFTW>`_.
+`here <http://pyFFTW.github.io/pyFFTW/>`_, and the source
+is on `github <https://github.com/pyFFTW/pyFFTW>`_.
 '''
 
 class custom_build_ext(build_ext):
@@ -460,7 +460,7 @@ def setup_package():
         'description': (
             'A pythonic wrapper around FFTW, the FFT library, presenting a '
             'unified interface for all the supported transforms.'),
-        'url': 'http://hgomersall.github.com/pyFFTW/',
+        'url': 'http://pyFFTW.github.io/pyFFTW/',
         'long_description': long_description,
         'classifiers': [
             'Programming Language :: Python',
--- a/README.rst
+++ b/README.rst
@@ -8,7 +8,7 @@
 
 .. |travis_ci| image:: https://travis-ci.org/pyFFTW/pyFFTW.svg?branch=master
    :align: middle
-   :target: https://travis-ci.org/hgomersall/pyFFTW
+   :target: https://travis-ci.org/pyFFTW/pyFFTW
 
 .. |appveyor_ci| image:: https://ci.appveyor.com/api/projects/status/uf854abck4x1qsjj/branch/master?svg=true
    :align: middle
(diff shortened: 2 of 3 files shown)


setup.py
```diff
@@ -216,8 +216,8 @@ def get_extensions():
 Numpy is a dependency for both.
 
 The documentation can be found
-`here <http://hgomersall.github.com/pyFFTW/>`_, and the source
-is on `github <https://github.com/hgomersall/pyFFTW>`_.
+`here <http://pyFFTW.github.io/pyFFTW/>`_, and the source
+is on `github <https://github.com/pyFFTW/pyFFTW>`_.
 '''
```
hgomersall: It actually _isn't_ on pyfftw.github.io. It should be, but I haven't got round to it yet. There are a few aspects of the docs that need sorting tbh (like autogenerating them on a release).

setup.py
```diff
@@ -460,7 +460,7 @@ def setup_package():
         'description': (
             'A pythonic wrapper around FFTW, the FFT library, presenting a '
             'unified interface for all the supported transforms.'),
-        'url': 'http://hgomersall.github.com/pyFFTW/',
+        'url': 'http://pyFFTW.github.io/pyFFTW/',
```
hgomersall: ditto

setup.py
jakirkham: FWIW there does appear to be docs at that URL, but maybe a bit out-of-date. Will revert these.

hgomersall: Oh my bad, I tried to navigate to the root which gave me a 404. Apologies for that. I don't know how up to date it is actually. It's a while since I did a release :disappointed: .

setup.py
hgomersall: ditto

setup.py
hgomersall: Keep them, they should point there.

jakirkham: FWIW the org one has docs for 0.10.1 whereas the user one has docs for 0.10.4.

hgomersall: Yeah, that's not surprising, and is indicative of the need for a better release process.

jakirkham: Yeah, I have one repo that automatically does this with Travis CI. Though I have largely switched to ReadTheDocs as it is just less effort to maintain.

ref: https://github.com/nanshe-org/nanshe/blob/9473837d3c8b978014c3f35c31c9f263f44d7549/.travis.yml#L87-L112

jakirkham:
So just to be clear, do you want me to switch back to the org repo docs?

setup.py
hgomersall: RTD is problematic with Cython (was my experience list time I tried this).

hgomersall: (hence the partial upload to RTD)

hgomersall:
Either is fine, but ultimately we should move to the org repo.

jakirkham:
It's pointed back at the org repo for the docs.

hgomersall:
:+1:

jakirkham:
Also have added PR ( https://github.com/pyFFTW/pyFFTW/pull/158 ), which should bring the docs in the org up to parity with those in your repo.

jakirkham:
The docs appear to now be up to date on the org after the merging of PR ( https://github.com/pyFFTW/pyFFTW/pull/158 ). So this should be good to go.

setup.py
jakirkham: Yeah, Sphinx needs to have the SOs built to actually read docs from them. So one is left to mock it out, which admittedly is a bit of an annoying task.

jakirkham: Well if you go the GitHub pages route, it might be worth getting it building on CI. Would recommend going with a dedicated tool instead of writing it out by hand like I did. Maybe something like [this]( https://github.com/Syntaf/travis-sphinx ).

hgomersall: I've half implemented a proper way of doing this, but ran out of time and never revisited it. It does need doing properly if you have the energy :smile:.

jakirkham:
Anything else needed for this one?

setup.py
jakirkham: I tried playing with RTD on the weekend, but didn't have any luck. It seems Sphinx needs to have the extensions built to get the docstrings, but RTD doesn't let one build the extensions. One option would be to mock out the Cython parts when building the docs. Not ideal though. 😞

grlee77: We are using RTD successfully along with Cython in `PyWavelets`.  I believe @rgommers was the one who set it up, so I don't have direct experience with it, but from what I can tell it looks like we have configured RTD to use a `pip requirements file` under the "Advanced Settings" tab of the the admin section on readthedocs.org.  This file points to the following requirements.txt that specifies that Cython should be installed:

https://github.com/PyWavelets/pywt/blob/master/util/readthedocs/requirements.txt

jakirkham: Thanks @grlee77. I did take a look at PyWavelets and a couple other projects when I was looking into RTD and Cython this weekend. Do you know if any of the docstrings from the Cython code is added to the documentation? I didn't seem to see any, but I could have easily missed it.

grlee77: I think some docstrings from Cython end up in the docs.  Here is one example I found for the `swt_max_level` routine which is defined in `pywt/_extensions/_swt.pyx` and then imported into `__all__` in  `pywt/_swt.py`.  It appears to be getting rendered correctly here:

http://pywavelets.readthedocs.io/en/latest/ref/swt-stationary-wavelet-transform.html#maximum-decomposition-level-swt-max-level

jakirkham: Just looked at PyWavelets' [most recent build log]( https://readthedocs.org/projects/pywavelets/builds/5129158/ ) and it seems that it is actually building the library as there is a `python setup.py install --force` step that occurs, which I don't have in my WIP branch for pyFFTW. Is there anything in PyWavelets beyond using a `requirements.txt` file that is causing this?

grlee77: It looks like there is also a checkbox to "Install Project" in the admin tab on ReadTheDocs.  This appears to be selected for PyWavelets:

screen shot 2017-04-03 at 6 29 16 pm

jakirkham: Wasn't able to do that exactly as it didn't work with `conda` (needed to install `fftw`), but did something similar. Now it seems to be unable to find `gcc`? Kind of strange. 😕 Log below if you are interested.

ref: https://readthedocs.org/projects/jakirkhampyfftw/builds/5243045/

jakirkham: Hacked around a bit and got something close to working in PR ( https://github.com/pyFFTW/pyFFTW/pull/161 ). However, I think have the docs mixed with the source code is causing problems, would propose we split them out into a separate `docs` (or otherwise named) directory separate from the source.

rgommers: I indeed set it up for PyWavelets, but I don't remember any particular problem with Cython that needed resolving beyond installing it via ``requirements.txt``. Certainly nothing like ``gcc`` being called ``x86_64-linux-gnu-gcc``. Sorry.

rgommers: splitting out ``docs`` is always a good idea for Sphinx based docs I think

jakirkham: Thanks @rgommers. Did some more tweaking this morning and got RTD to work. 🎉

Yeah the `gcc` name situation is strange. It seems PyWavelets does pick up the funky compiler name correctly, but pyFFTW didn't. Anyways ended up adding a hack that worked around this issue.

Splitting out the docs does still seem like a good idea, but it didn't fix the issue. Seems there were some Sphinx configuration issues that were the cause of the problem. Fixed them in PR ( https://github.com/pyFFTW/pyFFTW/pull/164 ).

hgomersall:
Easy I think. Thanks!

PR #157, opened 2017-03-30
Rename Cache interface tests

jakirkham:
Goes ahead and renames the cache interface test module to match the naming of the other interface test modules.


hgomersall:
This is a poor solution IMO. We should understand why the problem is occurring.

hgomersall:
I think this is fixed now in the original branch, correct?

jakirkham:
Correct, this is not needed as a workaround. Though it does match the naming of the other interfaces better.

hgomersall:
Hmm, it seems to have merged anyway (I must have missed it in another branch) ;)

jakirkham:
Oh, yeah, it's in the history of PR ( https://github.com/pyFFTW/pyFFTW/pull/154 ), but that PR also reverts this change. So this didn't change anything in the end. Would have rebased it out of the other change, but didn't want to rewrite the branch since you were working on it too. 😉

hgomersall:
It doesn't really matter. I think tbh I like the test name that maps to the module name, which is more like the cache case, but I really don't care that much.

PR #158, opened 2017-03-31
Update from hgomersall/gh-pages

jakirkham:
Pulls in all of the changes from `hgomersall/gh-pages` branch. This should bring the docs up-to-date here.


hgomersall:
Yup, thanks!

PR #159, opened 2017-03-31
add next_fast_len as introduced in scipy 0.18

### LINKED ISSUE #103: scipy_interfaces.fftpack needs _next_regular to support upstream changes
`_next_regular` is designed to support fft implementation specific optimisations. In the case of FFTW, this is different to fftpack, so a custom one should be included (it is used by scipy internal tools).

See https://github.com/scipy/scipy/pull/6124.

grlee77:
closes #103 

This introduces a routine as in `scipy.fftpack` to address the next fast favorable size for FFTs.  Unlike in scipy, FFTW supports additional prime factors.

From the FFTW3 documentation:

>FFTW is best at handling sizes of the form 2^a 3^b 5^c 7^d 11^e 13^f, where e+f is either 0 or 1, and the other exponents are arbitrary. Other sizes are computed by means of a slow, general-purpose algorithm (which nevertheless retains O(n log n) performance even for prime sizes). It is possible to customize FFTW for different array sizes; see Installation and Customization. Transforms whose sizes are powers of 2 are especially fast."

This same text is present in the descriptions for both real and complex FFTs:
http://fftw.org/fftw3_doc/Complex-DFTs.html#Complex-DFTs
http://fftw.org/fftw3_doc/Real_002ddata-DFTs.html#Real_002ddata-DFTs
http://fftw.org/fftw3_doc/Real_002dto_002dReal-Transforms.html#Real_002dto_002dReal-Transforms

I was not sure on the best location for this code.  I have placed it in `builders._utils` for now as its utility is not limited to `scipy.fftpack` wrappers.  I guess the function should also be imported into the `scipy.fftpack` interface namespace as `scipy` users would expect to find it there.

will follow up with a quick performance validation example
also still need to add it to the docs once we settle on the final location

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -29,7 +29,7 @@ autodoc_docstring_signature = True
 # Add any Sphinx extension module names here, as strings. They can be extensions
 # coming with Sphinx (named 'sphinx.ext.*') or your custom ones.
 extensions = ['sphinx.ext.autodoc', 'sphinx.ext.doctest', 'sphinx.ext.coverage',
-        'sphinx.ext.intersphinx']
+        'sphinx.ext.intersphinx', 'sphinx.ext.napoleon']
 
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ['_templates']
--- a/pyfftw/__init__.py
+++ b/pyfftw/__init__.py
@@ -27,6 +27,7 @@ from .pyfftw import (
         empty_aligned,
         ones_aligned,
         zeros_aligned,
+        next_fast_len,
 )
 
 from . import builders
--- a/pyfftw/interfaces/scipy_fftpack.py
+++ b/pyfftw/interfaces/scipy_fftpack.py
@@ -65,6 +65,9 @@ from scipy.fftpack import (dct, idct, diff, tilbert, itilbert,
         shift, fftshift, ifftshift, fftfreq, rfftfreq,
         convolve, _fftpack)
 
+# a next_fast_len specific to pyFFTW is used in place of the scipy.fftpack one
+from ..pyfftw import next_fast_len
+
 try:
     from scipy.fftpack import dst, idst
 except ImportError:
@@ -72,7 +75,8 @@ except ImportError:
 
 __all__ = ['fft','ifft','fftn','ifftn','rfft','irfft', 'fft2','ifft2',
         'diff', 'tilbert','itilbert','hilbert','ihilbert', 'sc_diff',
-        'cs_diff','cc_diff','ss_diff', 'shift', 'rfftfreq']
+        'cs_diff','cc_diff','ss_diff', 'shift', 'rfftfreq',
+        'next_fast_len']
 
 def fft(x, n=None, axis=-1, overwrite_x=False,
         planner_effort='FFTW_MEASURE', threads=1,
--- a/pyfftw/pyfftw.rst
+++ b/pyfftw/pyfftw.rst
@@ -92,3 +92,5 @@ Utility Functions
 .. autofunction:: pyfftw.n_byte_align_empty
 
 .. autofunction:: pyfftw.is_n_byte_aligned
+
+.. autofunction:: pyfftw.next_fast_len
--- a/pyfftw/utils.pxi
+++ b/pyfftw/utils.pxi
@@ -274,14 +274,14 @@ cpdef next_fast_len(target):
     >>> b = scipy_fftpack.fft(a)
 
     Zero-padding to the next fast length reduces computation time to
-    406 us, a speedup of >500 times:
+    406 us, a speedup of ~5 times:
 
-    >>> next_fast_len_pyfftw(min_len)
+    >>> next_fast_len(min_len)
     10080
     >>> b = scipy_fftpack.fft(a, 10080)
 
     Rounding up to the next power of 2 is not optimal, taking 598 us to
-    compute, 1.5 times as long as the size selected by next_fast_len_pyfftw
+    compute, 1.5 times as long as the size selected by next_fast_len.
 
     >>> b = fftpack.fft(a, 16384)
 
--- a/pyfftw/utils.pxi
+++ b/pyfftw/utils.pxi
@@ -35,6 +35,7 @@
 # POSSIBILITY OF SUCH DAMAGE.
 #
 
+from bisect import bisect_left
 cimport numpy as np
 cimport cpu
 from libc.stdint cimport intptr_t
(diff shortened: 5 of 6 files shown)


hgomersall:
This is great, thanks for doing this!

This should be in the same namespace as the equivalent in scipy, so it can drop in seamlessly. In `scipy`, it's in both `scipy.fftpack.helpers` as well as `scipy.fftpack` (to which I assume it is imported).

I think it is sufficient to have it only in `pyfftw.interfaces.scipy_fftpack`, which would mean that you could just include it in `interfaces/scipy_fftpack.py`. The tests can then be put in the equivalent file in `tests`.

endolith:
Yeah it's meant to be used like this https://hgomersall.github.io/pyFFTW/sphinx/tutorial.html#monkey-patching-3rd-party-libraries

https://github.com/scipy/scipy/pull/6124

(Note that variable name "hams" = "Hamming numbers", which yours are not)

grlee77:
> (Note that variable name "hams" = "Hamming numbers", which yours are not)

ahh... I was wondering what "hams" stood for.  can change the variable name here, then.

Related to this, do you feel it is important to test with **very** large integers as was done in the scipy tests?  I have not currently replicated those here (mainly I did not work out what the proper increment would be in the `large size + 1` cases)

>I think it is sufficient to have it only in pyfftw.interfaces.scipy_fftpack, which would mean that you could just include it in interfaces/scipy_fftpack.py. The tests can then be put in the equivalent file in tests.

I understand about having it in the `scipy` interface namespace, but I guess I was thinking that even if one is not using the scipy interface, the `next_fast_len` function will be of use in determining a good transform size.  I can move it where you suggest, but should we also import it at the top level so a user could do:
  `from pyfftw import next_fast_len`?

grlee77:
here is a basic benchmark result validating the current implementation.  (I have used the FFTW builders rather than the scipy interface)

```python
    import time
    import numpy as np
    from matplotlib import pyplot as plt
    from pyfftw.builders import fftn as build_fftn
    from pyfftw.builders._utils import next_fast_len

    all_l = np.arange(2, 1000)
    all_nl = np.zeros_like(all_l)
    all_times = np.zeros(len(all_l))
    all_times_pad = np.zeros(len(all_l))
    time_ratios = np.ones(len(all_l))
    for cnt, l in enumerate(all_l):
        l = int(l)
        print("l = {}".format(l))
        nl = next_fast_len(l)
        all_nl[cnt] = nl
        dl = np.random.randn(l)
        dnl = np.random.randn(nl)

        fft_l = build_fftn(dl, threads=1)
        fft_nl = build_fftn(dnl, threads=1)
        if l < 100:
            nreps = 10000
        elif l < 1000:
            nreps = 1000
        else:
            nreps = 100
        tstart = time.time()
        for n in range(nreps):
            fft_l(dl)
        tl = (time.time() - tstart) / nreps

        if l == nl:
            all_times[cnt] = tl
            all_times_pad[cnt] = tl
            time_ratios[cnt] == 1
            continue

        tstart = time.time()
        for n in range(nreps):
            fft_nl(dnl)
        tnl = (time.time() - tstart) / nreps

        all_times[cnt] = tl
        all_times_pad[cnt] = tnl
        time_ratios[cnt] = all_times_pad[cnt] / all_times[cnt]

    plt.figure(); plt.plot(all_l, 1e3*all_times, all_l, 1e3*all_times_pad)
    plt.legend(['unpadded', 'pad to next fast len'], loc='upper left')
    plt.xlabel('unpadded FFT length')
    plt.ylabel('time (ms)')
```
timings

or in terms of the acceleration factor:
ratios

endolith:
> Related to this, do you feel it is important to test with very large integers as was done in the scipy tests? I have not currently replicated those here (mainly I did not work out what the proper increment would be in the large size + 1 cases)

Probably not?  I went overboard with making sure it didn't miss any corner cases and optimizing for speed.  http://stackoverflow.com/a/19649737/125507  Then again maybe people do super-long FFTs on supercomputers? I don't know.

hgomersall:
Sure, `pyfftw.next_fast_len` works. Adding it to `utils.pxi` means it is exposed to cython as well which is useful. Just use `cpdef` rather than `def`, and add the signature on the first line, and it should work properly (not sure if there is a doc page that needs tweaking too, but it should be easy).

grlee77:
Okay.  I like the `utils.pxi` suggestion better, so I have moved it there.  I now import it from there into the scipy interfaces.  I have rebased/squashed into a single commit to get ride of all the unnecessary history of moving things around.

hgomersall:
Looks good, though I'd like to check it a little more carefully. Does it basically implement the scipy equivalent?

What does napoleon do?

grlee77:
The documentation I had for `next_fast_len` was using numpy style, but I did not realize that the `numpydoc` sphinx extension is not currently a dependency for pyFFTW.  A workaround that avoids adding it as a dependency is to use the  `sphinx.ext.napoleon` extension which does recognize the `Parameters`, `Returns`, and `Examples` fields.  (This extension has been included with sphinx itself by default since version 1.3).

Is this an acceptable solution, or how would you prefer to format the docstring for `next_fast_len`?

grlee77:
without the extension I get the following warnings when building the docs:
```
docstring of pyfftw.next_fast_len:7: SEVERE: Unexpected section title.

Parameters
----------
docstring of pyfftw.next_fast_len:12: SEVERE: Unexpected section title.

Returns
-------
docstring of pyfftw.next_fast_len:17: SEVERE: Unexpected section title.

Examples
--------
```

hgomersall:
Yeah, that's fine. The docs could do with an overhaul to standardise this sort of thing. What does it generate in the output?

grlee77:
Basically, napoleon recognizes both the numpy and google docstring styles.  There is a brief overview here:
https://pypi.python.org/pypi/sphinxcontrib-napoleon

grlee77:
It ends up rendering like this:
screen shot 2017-04-03 at 1 09 12 pm

hgomersall:
It looks good. I think it's how it _ought_ to be across the whole package.

hgomersall:
Is the implementation drawn from the scipy implementation?

grlee77:
Yes.  The core implementation for the inner loops of factors 2, 3 and 5 is exactly copied from scipy.  An additional loop is there for factor 7.  The factor 11 and 13 is handled slightly differently so there is only at most a factor of 11 or 13 but not both as described in the docstring.  There are a couple of cases explictly for this case.

The variable at the top is basically just precomputed results for short transform lengths so that the overhead of computing the factors can be avoided in these cases.  This list was updated to match the pyFFTW factors and is thus longer than the one present in scipy.

hgomersall:
Ok, great. I'll take a look later on, but I suspect it's fine. Thanks!

grlee77:
for reference, the `scipy` implementation is here:

https://github.com/scipy/scipy/blob/ab9eff32ef594a59621404ffe18045fd6bd49e1f/scipy/fftpack/helper.py#L51

hgomersall:
This is great, thanks!

PR #160, opened 2017-04-03
ENH: add norm keyword argument to the rfft*, hfft* numpy interfaces

### LINKED ISSUE #152: Missing norm argument in several numpy interface fft methods
In the [pyfftw numpy interfaces module](https://github.com/pyFFTW/pyFFTW/blob/master/pyfftw/interfaces/numpy_fft.py), some of the fft methods do not have a `norm` argument. For instance: 
```
def fft(a, n=None, axis=-1, norm=None, overwrite_input=False,
        planner_effort='FFTW_ESTIMATE', threads=1,
        auto_align_input=True, auto_contiguous=True)
```

has a `norm` argument, whereas 

```
def rfft(a, n=None, axis=-1, overwrite_input=False,
        planner_effort='FFTW_ESTIMATE', threads=1,
        auto_align_input=True, auto_contiguous=True):
```

does not.

The [numpy version](https://docs.scipy.org/doc/numpy/reference/generated/numpy.fft.rfft.html#numpy.fft.rfft) of the`rfft` method, does have a `norm` argument, as do all other numpy `*fft*` methods.

grlee77:
This PR adds the `norm` keyword to the remaining numpy fft interfaces.  A new `_norm_kwargs` function was defined to avoid so many duplicates of the following code across functions:

```python
if _unitary(norm):
    ortho = True
    normalise_idft = False
else:
    ortho = False
    normalise_idft = True
```

One tricky aspect is related to a bug in numpy's `rfft` when `n != None` that was only recently fixed in numpy/numpy#8445.  When necessary (i.e. numpy < 1.13), I have monkey-patched the `rfft` and `rfft2` and `rfftn` routines used for validation in the tests so that the tests are performed against these corrected versions.  Without this patching, a handful of the RFFT tests will fail due to the `numpy` bug.

closes #152

### DIFF
--- a/pyfftw/interfaces/numpy_fft.py
+++ b/pyfftw/interfaces/numpy_fft.py
@@ -74,6 +74,16 @@ def _unitary(norm):
                          % norm)
     return norm is not None
 
+def _norm_args(norm):
+    """ pass the proper normalization-related keyword arguments. """
+    if _unitary(norm):
+        ortho = True
+        normalise_idft = False
+    else:
+        ortho = False
+        normalise_idft = True
+    return dict(normalise_idft=normalise_idft, ortho=ortho)
+
 def fft(a, n=None, axis=-1, norm=None, overwrite_input=False,
         planner_effort='FFTW_ESTIMATE', threads=1,
         auto_align_input=True, auto_contiguous=True):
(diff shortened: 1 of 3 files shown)


hgomersall:
Looks great, thanks!

timseries:
@grlee77 Awesome, thanks for taking care of this!

PR #161, opened 2017-04-04
Configure RTD

jakirkham:
Builds the docs for pyFFTW on ReadTheDocs. Requires a hack to fix use the same toolchain that ReadTheDocs uses. Also relies on PR ( https://github.com/pyFFTW/pyFFTW/pull/162 ) to set some reasonable search paths for use with `conda` packages and PR ( https://github.com/pyFFTW/pyFFTW/pull/164 ) to fix some things with the Sphinx configuration.

ref: http://jakirkhampyfftw.readthedocs.io/en/config_rtd/index.html
ref: https://readthedocs.org/projects/jakirkhampyfftw/builds/5245737

cc @hgomersall @grlee77

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -16,7 +16,7 @@ import sys, os
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
-sys.path.insert(0, os.path.abspath('.'))
+#sys.path.insert(0, os.path.abspath('.'))
 
 # -- General configuration -----------------------------------------------------
 
--- a/conf.py
+++ b/conf.py
@@ -51,7 +51,7 @@ copyright = u'2016, Henry Gomersall'
 # |version| and |release|, also used in various other places throughout the
 # built documents.
 #
-from setup import VERSION as pyfftw_version
+from pyfftw.version import version as pyfftw_version
 # The short X.Y version.
 version = pyfftw_version
 # The full version, including alpha/beta/rc tags.
--- a/environment_doc.yml
+++ b/environment_doc.yml
@@ -1,0 +1,13 @@
+name: pyfftw_doc_env
+
+channels:
+  - conda-forge
+
+dependencies:
+  - cython
+  - fftw
+  - numpy
+  - python
+  - scipy
+  - dask
+  - sphinx
--- a/readthedocs.yml
+++ b/readthedocs.yml
@@ -1,0 +1,4 @@
+conda:
+  file: environment_doc.yml
+python:
+  setup_py_install: true
--- a/setup.py
+++ b/setup.py
@@ -81,6 +81,18 @@ ISRELEASED = True
 
 VERSION = '%d.%d.%d' % (MAJOR, MINOR, MICRO)
 
+
+if os.environ.get("READTHEDOCS") == "True":
+    try:
+        environ = os.environb
+    except AttributeError:
+        environ = os.environ
+
+    environ[b"CC"] = b"x86_64-linux-gnu-gcc"
+    environ[b"LD"] = b"x86_64-linux-gnu-ld"
+    environ[b"AR"] = b"x86_64-linux-gnu-ar"
+
+
 def get_package_data():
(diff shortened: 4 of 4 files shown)


jakirkham:
Tried restructuring the docs, but that doesn't seem to have fixed it. It seems like there is still something off. FWICT it seems to be related to the import structure somehow. 😕

jakirkham:
🎉 This now works! 🎉

Please take a look and share feedback.

grlee77:
great to see this is now working!

I was not previously familiar with the `readthedocs.yml` for specifying conda packages.  That is pretty nice!

It's not ideal that we have to hack `setup.py` to find `gcc`, but I don't see that it should cause any problems.  I guess an alternative hack would be to try adding the `gcc` package from conda-forge in the `environment.yml` and see if that one gets picked up by default? (at cost of slightly longer build times do to the additional gcc download).

jakirkham:
Thanks for the feedback, @grlee77.

Yeah, `conda` support is relatively new. It had been in the works for a while, but got completed last year. For a bit of history, please see issue ( https://github.com/rtfd/readthedocs.org/issues/857 ).

Using `gcc` is a good idea. I tried this initially, but it didn't work. The `gcc` package in use has some post link script, which [failed]( https://readthedocs.org/projects/jakirkhampyfftw/builds/5243017 ) in the RTD build environment. This post link script expects to be able to find out some information about the environment and makes a lot of assumption based on a variety of Linux distros tried. However it just is unable to do this for the RTD build environment for whatever reason. I tried rolling back to earlier versions of the `gcc` package including one without this post link script, but that [fails]( https://readthedocs.org/projects/jakirkhampyfftw/builds/5243042/ ) due to the short prefix. Given these issues, I decided trying to change the compiler used in `setup.py` for RTD was the least bad option. Still not sure why it is necessary for pyFFTW, but not necessary for PyWavelets.

hgomersall:
I'll have to look over this later, I need to run now.

Are there any potential issues with these changes? Or is the net effect that the RTD docs are up to date?

jakirkham:
It has PR ( https://github.com/pyFFTW/pyFFTW/pull/164 ) in the history. So might want to check if you are ok with that change.

Otherwise, no, it just sets up RTD to build.

hgomersall:
Apologies for the delay on this, I've been very busy.

jakirkham:
No worries. I think the RTD project settings might need to be refreshed though. Unfortunately there is not much I can do on from my end.

hgomersall:
What needs to be done? When you say not much can be done is that because I need to do it?

jakirkham:
Yes. It's been a little while since I did this, but I would first check to see if the RTD service is enabled on this repo. Maybe it was disabled after the initial setbacks, which makes sense.

I tried to put everything else in the `readthedocs.yml` file in this PR. So I don't think there are any setting changes. However if it doesn't work after enabling RTD and doing a fresh rebuild, please ping me and I can look closer.

hgomersall:
Ok will do, thanks. btw, I invited you to be a maintainer. It's about time I excited other people in pyFFTW and you've been keen enough so far!

jakirkham:
So I tried creating a fresh RTD project without any setting changes and that seemed to work. It built a copy of docs under `latest` correctly. `stable` was not correct, but I think that uses the latest tag instead of `master`, which is what `latest` does. Guessing the only thing that needs to be updated is the webhook then.

jakirkham:
Cool, thanks. I plan to extend the Dask interface to 2-D and N-D on the next release of Dask.

We like using pyFFTW internally for image registration. It makes for a nice speedup beyond what NumPy' FFTPACK can do. Not to mention if it helps a convolution with a large kernel go a bit faster, that is nice too.

hgomersall:
So I've set up the webhooks. I guess we see what happens...

For the time being, I'm happy with just latest working!

hgomersall:
I really appreciate all your work on this, it's very valuable.

jakirkham:
Great, thanks. No problem. FWIW it looks like it is working.

FYI may need to clear your cache if the docs aren't showing up correctly.

hgomersall:
No, it looks fine :)

PR #162, opened 2017-04-04
Check system prefix for headers and libraries

jakirkham:
Uses `sys.prefix` as a root search path for headers and libraries. This is where `conda` stores everything relative to. So it should simplify building `conda` packages or building in a `conda` environment with other build tools (e.g. `pip`).

cc @grlee77


hgomersall:
I think this is safe... Thanks.

PR #164, opened 2017-04-04
Fix sphinx configuration

jakirkham:
Fixes a couple of things with the configuration file that came up while trying to get ReadTheDocs working.

First it stops adding the current working directory to the path. This ends up resulting in some confusion for Sphinx (with or without ReadTheDocs) in terms of where to import things from. By dropping this path manipulation, Sphinx simply tries to import `pyfftw` the standard way. So if it was installed in `site-packages`, this will import it from there. Also it works just as well with development installs. Regardless it ensures it is importing only the fully built `pyfftw` and not the unbuilt source, which causes problems.

Second it gets the version information from `pyfftw.version`, which should exist in the fully built `pyfftw` instead of resorting to using the `setup.py` file. Importing the latter could have all sorts of weird, unintended side effects. The same problems are not true of `pyfftw.version`. Thus it works just as well and is cleaner.

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -16,7 +16,7 @@ import sys, os
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
-sys.path.insert(0, os.path.abspath('.'))
+#sys.path.insert(0, os.path.abspath('.'))
 
 # -- General configuration -----------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


conf.py
```diff
@@ -16,7 +16,7 @@
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
-sys.path.insert(0, os.path.abspath('.'))
+#sys.path.insert(0, os.path.abspath('.'))
```
jakirkham: If we don't like this, we can add a guard that checks the [`READTHEDOCS` environment variable]( http://docs.readthedocs.io/en/latest/faq.html#how-do-i-change-behavior-for-read-the-docs ).

jakirkham:
Please let me know your thoughts on this, @hgomersall.

PR #165, opened 2017-04-04
Import functions directly from the NumPy FFT interface

jakirkham:
Instead of importing the NumPy FFT module interface, import functions from that interface and then wrap those functions. This avoids things leaking in from the NumPy FFT interface into the Dask FFT interface.

### DIFF
--- a/pyfftw/interfaces/dask_fft.py
+++ b/pyfftw/interfaces/dask_fft.py
@@ -53,14 +53,21 @@ equivalents in :mod:`dask.fft`, though there are some corner cases in
 which this may not be true.
 '''
 
-from . import numpy_fft
+from .numpy_fft import (
+    fft,
+    ifft,
+    rfft,
+    irfft,
+    hfft,
+    ihfft,
+)
 from dask.array.fft import (
     fft_wrap,
 )
 
-fft = fft_wrap(numpy_fft.fft)
-ifft = fft_wrap(numpy_fft.ifft)
-rfft = fft_wrap(numpy_fft.rfft)
-irfft = fft_wrap(numpy_fft.irfft)
-hfft = fft_wrap(numpy_fft.hfft)
-ihfft = fft_wrap(numpy_fft.ihfft)
+fft = fft_wrap(fft)
+ifft = fft_wrap(ifft)
+rfft = fft_wrap(rfft)
+irfft = fft_wrap(irfft)
+hfft = fft_wrap(hfft)
+ihfft = fft_wrap(ihfft)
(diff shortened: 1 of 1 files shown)


jakirkham:
Thoughts on this, @hgomersall?

hgomersall:
Sorry, I've been really busy this last week.

What's the point of this? I'm not a big fan of overwriting the namespace like you do. What was wrong with the previous way?

jakirkham:
Because we shouldn't have `numpy_fft` available in `dask_fft`.

jakirkham:
> I'm not a big fan of overwriting the namespace like you do.

Also I don't follow this. How are we overwriting the namespace? We are only importing select functions from `numpy_fft` instead of importing the whole module. What is wrong with doing it this way?

pyfftw/interfaces/dask_fft.py
```diff
@@ -53,14 +53,21 @@
 which this may not be true.
 '''
 
-from . import numpy_fft
+from .numpy_fft import (
+    fft,
```
hgomersall: Here `dask_fft.fft` is the same as `numpy_fft.fft`

pyfftw/interfaces/dask_fft.py
```diff
@@ -53,14 +53,21 @@
...
     fft_wrap,
 )
 
-fft = fft_wrap(numpy_fft.fft)
-ifft = fft_wrap(numpy_fft.ifft)
-rfft = fft_wrap(numpy_fft.rfft)
-irfft = fft_wrap(numpy_fft.irfft)
-hfft = fft_wrap(numpy_fft.hfft)
-ihfft = fft_wrap(numpy_fft.ihfft)
+fft = fft_wrap(fft)
```
hgomersall: Here it changes.

jakirkham:
Sure, that is intentional. This is not in any way different from how Python decorators work.

Though maybe I'm not understanding your concern. Could you please clarify what problem you see this causing?

hgomersall:
It's a bit hypothetical, but imported code could change on the fly (particularly if we consider threads). It's not quite how decorators work - they don't put anything in the namespace until after the decorator returns (or rather, the object put into the namespace always has the correct getter associated with it).

hgomersall:
Actually, I don't know how what I said could happen. It's probably safe :) I'm not keen on the style though because it's not clear that the import fft is different to the one in the ultimate module namespace.

hgomersall:
I'd prefer it I think if you kept the existing implementation. You could "protect" `numpy_fft` by importing it with an underscore: `from . import numpy_fft as _numpy_fft`. That would mean it wouldn't import with `import *` nor come up with tab completion in ipython.

hgomersall:
*poke* (cheekily, given by slow response time).

hgomersall:
Great, thanks! I'll merge after successful tests.

jakirkham:
Thanks @hgomersall. Wasn't going to ping you until it passed, but I guess it just did.

PR #166, opened 2017-04-04
Make sure to reference Dask interface

jakirkham:
Accidentally was referencing part of the NumPy FFT interface again instead of referencing the Dask FFT interface. Updated to reference the appropriate interface.

### DIFF
--- a/pyfftw/interfaces/__init__.py
+++ b/pyfftw/interfaces/__init__.py
@@ -128,15 +128,15 @@ being taken along that axes as many times as the axis occurs.
 * :func:`pyfftw.interfaces.scipy_fftpack.rfft`
 * :func:`pyfftw.interfaces.scipy_fftpack.irfft`
 
-:mod:`~pyfftw.interfaces.numpy_fft`
+:mod:`~pyfftw.interfaces.dask_fft`
 """""""""""""""""""""""""""""""""""
 
-* :func:`pyfftw.interfaces.numpy_fft.fft`
-* :func:`pyfftw.interfaces.numpy_fft.ifft`
-* :func:`pyfftw.interfaces.numpy_fft.rfft`
-* :func:`pyfftw.interfaces.numpy_fft.irfft`
-* :func:`pyfftw.interfaces.numpy_fft.hfft`
-* :func:`pyfftw.interfaces.numpy_fft.ihfft`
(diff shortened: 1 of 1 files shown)


grlee77:
LGTM.

p.s.  there is a warning when building the docs if `dask` is not installed, but it doesn't seem to stop the build from completing.  I think the same would occur for the scipy interfaces if scipy isn't installed, so probably not worth worrying about.  clearly we can't document the routines if they aren't present

```
/Users/lee8rx/src/my_git/pyFFTW/pyfftw/interfaces/dask_fft.rst:4: WARNING: autodoc: failed to import module 'pyfftw.interfaces.dask_fft'; the following exception was raised:
  File "/Users/lee8rx/anaconda/lib/python3.5/site-packages/Sphinx-1.5.1-py3.5.egg/sphinx/ext/autodoc.py", line 547, in import_object
    __import__(self.modname)
  File "/Users/lee8rx/src/my_git/pyFFTW/pyfftw/interfaces/dask_fft.py", line 57, in <module>
    from dask.array.fft import (
ImportError: cannot import name 'fft_wrap'
```

jakirkham:
Yeah, we decided in PR ( https://github.com/pyFFTW/pyFFTW/pull/154 ) to keep Dask as an optional dependency. So the warning is not surprising. As you note, not much we can do about it.

FWIW with ReadTheDocs we make sure to include Dask in PR ( https://github.com/pyFFTW/pyFFTW/pull/161 ). So it should be included and documented. In fact, I can find the Dask interface in the [docs]( http://jakirkhampyfftw.readthedocs.io/en/config_rtd/pyfftw/interfaces/dask_fft.html ) and it looks correct. Just this listing was incorrect by not referencing it.

**Side Note:** The Dask interface was designed to follow the SciPy interface pretty closely in terms of being optional. IOW importing `pyfftw.interfaces` will try to import `dask` and populate the `dask_fft` interface. However, if the import of `dask` doesn't work or the version of `dask` is too old, then we clean up all traces of `dask` and the associated `dask_fft` interface. So importing `pyfftw.interfaces` will always work, but it may just not include certain interfaces if the required dependencies and versions are missing.

jakirkham:
Thoughts @hgomersall?

hgomersall:
Looks good. It's not unreasonable to need all the deps to build the docs.

PR #163, opened 2017-04-04
Move docs out of the source code

jakirkham:
Moves the source, configuration files, and build script for the docs into the `docs` directory. Also updates all paths relevant paths based on this new location.

PR #167, opened 2017-04-05
extend the norm argument as implemented in the numpy interface to the builders

grlee77:
This allows the FFTW builders to all take either `norm=None` (matches the current behaviour) or `norm='ortho'`in exactly the same way as the numpy interfaces.

In implementing, this two of the arguments from the `__call__` method of the core FFTW class were also added to the `__cinit__` method.  These are the `ortho` and `normalise_idft` arguments related to normalisation.  This allows the default normalisation to be set at the time of class initialization, but this choice can still be overridden by passing explicit values to the `__call__` method if desired.

The private convenience functions for normalisation argument handling ( `_norm_args` and `_unitary`) were moved from `pyfftw.interfaces.numpy_fft.py` to `pyfftw.builders._utils.py` as they are now used by both the builder and numpy interface routines.

### DIFF
--- a/pyfftw/pyfftw.pyx
+++ b/pyfftw/pyfftw.pyx
@@ -833,7 +833,7 @@ cdef class FFTW:
     def __cinit__(self, input_array, output_array, axes=(-1,),
                   direction='FFTW_FORWARD', flags=('FFTW_MEASURE',),
                   unsigned int threads=1, planning_timelimit=None,
-                  normalise_idft=True, ortho=False,
+                  bint normalise_idft=True, bint ortho=False,
                   *args, **kwargs):
 
         # Initialise the pointers that need to be freed
--- a/pyfftw/pyfftw.pyx
+++ b/pyfftw/pyfftw.pyx
@@ -815,8 +815,7 @@ cdef class FFTW:
 
     def _get_normalise_idft(self):
         '''
-        If normalise_idft, the inverse transform is scaled by 1/N.
-        N is the product of the shape along the transformed axes.
+        If ``normalise_idft=True``, the inverse transform is scaled by 1/N.
         '''
         return self._normalise_idft
 
@@ -824,7 +823,7 @@ cdef class FFTW:
 
     def _get_ortho(self):
         '''
-        If ortho both the forward and inverse transforms are scaled by
+        If ``ortho=True`` both the forward and inverse transforms are scaled by
         1/sqrt(N).
         '''
         return self._ortho
--- a/pyfftw/pyfftw.pyx
+++ b/pyfftw/pyfftw.pyx
@@ -848,7 +848,7 @@ cdef class FFTW:
         self._normalise_idft = normalise_idft
         self._ortho = ortho
         if self._ortho and self._normalise_idft:
-            raise ValueError('Incompatible normalization options: '
+            raise ValueError('Invalid options: '
                 'ortho and normalise_idft cannot both be True.')
 
         flags = list(flags)
--- a/pyfftw/pyfftw.pyx
+++ b/pyfftw/pyfftw.pyx
@@ -665,6 +665,10 @@ cdef class FFTW:
     cdef int64_t *_not_axes
 
     cdef int64_t _N
+
+    cdef bint _normalise_idft
+    cdef bint _ortho
+
     def _get_N(self):
         '''
         The product of the lengths of the DFT over all DFT axes.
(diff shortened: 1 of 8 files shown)


pyfftw/pyfftw.pyx
```diff
@@ -1408,6 +1436,11 @@ cdef class FFTW:
...
         '''
 
+        if ortho is None:
+            ortho = self._ortho
+        if normalise_idft is None:
+            normalise_idft = self._normalise_idft
+
         if ortho and normalise_idft:
             raise ValueError('Invalid options: ortho and normalise_idft cannot'
                              ' both be True.')
```
hgomersall: I think this test should be more explicit. I don't much like `None` being used as `False`.

hgomersall:
Does this fix a real problem? Is it mostly to make the API more closely matched to Numpy?

Do you expect this to work with no problems at all? I can't think of any, but I'm a little wary because of how far reaching it could be.

pyfftw/pyfftw.pyx
grlee77: Here, `None` does **not** imply `False`.  It implies using whatever value for `ortho` or `_normalise_idft` were set to in the `__cinit__` constructor.  This allows one to create an FFTW object with the desired normalisation behavior.  Once the object has been created, it is never necessary to pass either kwarg to the `__call__` method unless it is desired to override the choices made at the time of the FFTW object creation.

Perhaps there is a better way to achieve the same goal?  It might be a bit clearer to just remove the `ortho` and `_normalise_idft` kwargs from the `__call__` method, but that would involve substantially more refactoring elsewhere in the codebase, as these are occasionally used (e.g. in the numpy interfaces)

grlee77:
The goal here is to have a way for the builder routines to plan an FFT with 1/sqrt(N) normalization on both the foward and inverse transforms.  This is the form of normalization I most often use in applications and I would like to be able to use the builders rather than the numpy interfaces to avoid planning overhead.

I care less that the `norm` argument handling in the builders exactly matches that in `numpy`, but I figured that was a choice that users may already be familiar with.  (It also makes it a bit simpler to validate vs. the numpy routines in the tests).

In `numpy` there are only two options for normalisation:

- 1/sqrt(N) on both forward and inverse transforms  (if `norm='ortho'` then the FFTW object is initialized with `ortho=True`, `normalise_idft=False`)  **THIS PR ENABLES THIS OPTION IN THE BUILDERS**
- 1/N only on the inverse transform  (if `norm=None` is specified, the FFTW object is initialized with `ortho=False`, `normalise_idft=True`)  **THIS IS THE ONLY CURRENT OPTION IN THE BUILDERS**

while there is also one other possibility when using the FFTW objects that are not possible in numpy FFTs:
- do not normalize at all. (`ortho=False, normalise_idft=False`).  This PR has not added that last, entirely unnormalized case to the Builders.

A fourth possibility: 
- `ortho=True`, `normalise_idft=True` makes no sense and already resulted in a `ValueError`.

pyfftw/pyfftw.pyx
hgomersall: It's line 1444 which is the problem, not the lines above. Basically, lines 1444-6 need tweaking to match the new signature.

hgomersall:
This all sounds fine to me and I think it's a reasonable solution (not perfect, but that is down to prior design decisions). I'll review it shortly...

pyfftw/pyfftw.pyx
grlee77: I am still not sure I see the problem.  After 1439-1442 `ortho` and `normalise_idft` will both be either `True` or `False` as they will now have the values from the `__cinit__` constructor which requires bools (I just updated the `__cinit__` parameters to have type `bint` which I had previously neglected to do).

Please suggest a specific fix if I am still missing the point here

hgomersall: Oh my bad, in my mind I had `ortho` as `norm`. Ok, I think it works fine.

hgomersall:
Ach, conflicts. Would you mind moving `pyfftw/pyfftw.rst` to reflect the changes in #163 ? Apologies for taking a while to sort this, I've had a huge amount on.

It's probably time to assign additional rights to pyFFTW if you're interested?

grlee77:
okay.  it should be good to go now assuming all tests still pass after the rebase.

>It's probably time to assign additional rights to pyFFTW if you're interested?
Sure, that would be great.

PR #169, opened 2017-04-22
Documentation is on Read the Docs

jakirkham:
Updates all info about documentation to point to Read the Docs.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -231,7 +231,7 @@ This should probably work for OSX, though I've not tried it.
 Numpy is a dependency for both.
 
 The documentation can be found
-`here <http://pyFFTW.github.io/pyFFTW/>`_, and the source
+`here <http://pyfftw.readthedocs.io>`_, and the source
 is on `github <https://github.com/pyFFTW/pyFFTW>`_.
 '''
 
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -475,7 +475,7 @@ def setup_package():
         'description': (
             'A pythonic wrapper around FFTW, the FFT library, presenting a '
             'unified interface for all the supported transforms.'),
-        'url': 'http://pyFFTW.github.io/pyFFTW/',
+        'url': 'https://github.com/pyFFTW/pyFFTW',
```
jakirkham: The alternative to this would be to link to Read the Docs here as well.

jakirkham:
Would add that I think we should also update the link in the repo description too. Perhaps it should point to Read the Docs also. Thoughts?

hgomersall:
This looks great, except I'm not sure we have all the versions set up properly? Does RTD build for each branch? In which case we should branch off 0.10.4 (or whatever is on pypi currently) and build for that too (as well as future releases).

hgomersall:
This is easy to merge. Thanks.

PR #170, opened 2017-04-22
Redirect to Read the Docs

jakirkham:
Removes the documentation from GitHub Pages and adds a redirect to Read the Docs.

The redirection code added to the index pulls from several pieces of advice found in this [SO answer]( http://stackoverflow.com/q/5411538 ). First we try to redirect users to Read the Docs via HTML. Second we try to use JavaScript redirect as a fallback. Third we provide some text explaining where to go with a link if redirection fails. In addition to the redirections, we note the canonical link is Read the Docs for web crawlers, which should hopefully fix search engine results over time.

Also have pushed these to changes to the `gh-pages` branch on my fork. So changes should be visible from ( http://jakirkham.github.io/pyFFTW ).


CHANGELOG.md
```diff
@@ -1,146 +0,0 @@
-# Change Log
```
jakirkham: Also saw we have this outdated changelog. Do we want to migrate this into `master` and refresh it a bit?

jakirkham:
Please let me know your thoughts on this @hgomersall. I know this is a pretty sweeping and opinionated change. So would not want to go through with this unless it is ok with you.

CHANGELOG.md
hgomersall: Yes, we do! Ideally, it should be automatic (it was generated I think with [github-changelog-generator](https://github.com/skywinder/github-changelog-generator)).

hgomersall:
I thought I commented on this... hmmm.

We should do this. The question I have is regards whether we need a docs branch on RTD for the pypi release. Can we do this?

The changelog needs to be updated too. Can this be automated somehow?

jakirkham:
Sorry for taking a while to respond. Got pulled into other things.

> We should do this. The question I have is regards whether we need a docs branch on RTD for the pypi release.

Was actually hoping we could talk about doing a release soonish. Wanted to try and add some more stuff to the Dask interface first. Not sure if you or anyone else had outstanding stuff.

> Can we do this?

We could try. RTD let's one build different branches and/or tags if they would like. The only catch is it will need all the changes that we added to master to get RTD to work there. Not sure whether we want to do that or if we would rather hold off on this until the next release. Am ok with either really.

hgomersall:
Yeah, a release would be good. Given your sterling work on various aspects of integration, do you fancy doing some stuff on easing the release process? There are various tools people have used. I played around with one but didn't quite get it working as desired (basically, the goal would be to run one command which builds the changelog, releases a tagged commit, pushes to PyPI and bumps the version number, plus anything else necessary).

grlee77:
> The only catch is it will need all the changes that we added to master to get RTD to work there. Not sure whether we want to do that or if we would rather hold off on this until the next release.

@jakirkham.  not sure I fully understand.  Are you saying it may be easier to do a release and then merge this shortly afterwards?

hgomersall:
I'm happy with this. Thanks!

jakirkham:
Sorry I have been quiet here of late.

>> The only catch is it will need all the changes that we added to master to get RTD to work there. Not sure whether we want to do that or if we would rather hold off on this until the next release.
>
> not sure I fully understand. Are you saying it may be easier to do a release and then merge this shortly afterwards?

Was trying to say this would be hard to do for older releases.

PR #172, opened 2017-04-26
 Attempted fix to the problem of the FFTW libs not being downloaded on appveyor.

hgomersall:


hgomersall:
Ach, it's broken in a new way now. Windows is such a pain in the arse to get right.

grlee77:
Shouldn't Appveyor also run on PRs?  I am not sure where to set that, but it would help identify these problems sooner.

hgomersall:
Yes, it [runs and fails](https://ci.appveyor.com/project/hgomersall/pyfftw/build/job/04djomrrcky13si0). It's not feeding back to github at the moment though.

grlee77:
Looks like we could remove that miniconda install script altogether as it now comes pre-installed on Appveyor:
https://www.appveyor.com/docs/build-environment/#python
http://tjelvarolsson.com/blog/how-to-continuously-test-your-python-code-on-windows-using-appveyor/

hgomersall:
Great, let's try that...

hgomersall:
That seems to be working now... Thanks!

grlee77:
Great!  Looks good to me

hgomersall:
3.6 is broken. Still I think we can merge with what we have (better than nothing).

grlee77:
see #173 for an attempt at fixing Python 3.6.  The problem appears to be that numpy 1.10 is the oldest version for which there is a conda package for Python 3.6.

PR #173, opened 2017-04-26
BLD: no conda package for numpy1.9 on Python3.6 so bump to 1.10.4

grlee77:
bump NumPy up to 1.10 on Appveyor so a conda package will exist for Python 3.6

Note:  as implemented this changes all versions to build against 1.10, but could be modified to keep the rest at 1.9 if that is preferred.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -6,6 +6,7 @@ environment:
         CMD_IN_ENV: "cmd /E:ON /V:ON /C .\\appveyor\\run_with_env.cmd"
         # The miniconda installations are put in pyfftw_miniconda
         PYTHON_HOME: "C:\\pyfftw_miniconda"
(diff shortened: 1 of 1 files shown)


hgomersall:
I think 1.10 is old enough now. Do you have any idea what happens if you use a 1.10 build with >1.10?

grlee77:
I believe it is guaranteed to work with any newer numpy versions, but may not work with older numpy versions.  Actually, I misremembered and think 3.6 may need need 1.11 (see relevant discussion here):
https://github.com/MacPython/pywavelets-wheels/pull/1

grlee77:
I modified to use 1.11 on Python 3.6 and left the rest at 1.9

hgomersall:
Fantastic - all working! :smile:

Issue #70, opened 2015-10-16
Support not having all varieties of FFTW installed


coderforlife:
In some Linux distributions it is common to have only the fftw3 library and not fftw3f. PyFFTW should detect that it is not available and work anyways (just not accept float32/single-precision inputs). Instead it just fails to import the whole module.

(If you are wondering why the whole devel package is not available, it is because we have numerous machines set up with mostly the same setup, but only one machine with devel, then we can use any machine with those libraries).

rabernat:
:+1: 

I found this to be quite a pain. To deal with it, in the [pyqg docs](http://pyqg.readthedocs.org/en/stable/installation.html#the-hard-way-installing-from-source) we recommend building FFTW three separate times. This is the only way to make pyfftw happy.

hgomersall:
Three separate times? How did that work?

rabernat:
The only way to get all three different FFTW datatypes required for pyfftw is to do
```bash
$ ./configure --enable-threads --enable-shared
$ make
$ sudo make install
$ ./configure --enable-threads --enable-shared --enable-float
$ make
$ sudo make install
$ ./configure --enable-threads --enable-shared --enable-long-double
$ make
$ sudo make install
```

hgomersall:
Ah OK, that makes sense :)

On 5 November 2015 13:10:08 GMT+00:00, Ryan Abernathey <notifications@github.com> wrote:
>The only way to get all three different FFTW datatypes required for
>pyfftw is to do
>```bash
>$ ./configure --enable-threads --enable-shared
>$ make
>$ sudo make install
>$ ./configure --enable-threads --enable-shared --enable-float
>$ make
>$ sudo make install
>$ ./configure --enable-threads --enable-shared --enable-long-double
>$ make
>$ sudo make install
>```
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/hgomersall/pyFFTW/issues/70#issuecomment-154055809

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

ghisvail:
Same on Debian. Some architectures do not have FFTW compiled with long-double.

One solution could be to use `pkgconfig` to detect whether each precision is available.

hgomersall:
I'm keen for help on this. Some thought it needed as to how to implement and test this properly, so I'd welcome any input.

ghisvail:
This is the patch I currently apply on Debian to build on architectures which do not have all versions of FFTW:
[...]

PR #67, opened 2015-09-19
Support FFTW + MPI

fredRos:
This addresses #38

Issue #138, opened 2016-12-05
importing numpy before pyfftw breaks wisdom / planning (really?!)


zpincus:
This seems completely bizarre (and took forever to isolate), but it appears that importing numpy before pyfftw breaks things regarding wisdom and planning. 

Here's a simple test function:
```python
def plan():
    image_arr = pyfftw.empty_aligned(shape=(2560,2160), dtype=numpy.float32, order='F')
    fft_arr = pyfftw.empty_aligned(shape=(2560, 1081), dtype=numpy.complex64, order='F')
    return pyfftw.FFTW(image_arr, fft_arr, axes=(0,1), direction='FFTW_FORWARD', flags=('FFTW_MEASURE',))
```

If I run the following, everything works as expected:
```python
import pyfftw
import numpy

print(pyfftw.export_wisdom()) # not much here
plan() # takes a second or two
print(pyfftw.export_wisdom()) # lots of wisdom
```

**However**, if I merely transpose the two import statements above, things work very differently:
```python
import numpy
import pyfftw

print(pyfftw.export_wisdom()) # some empty strings??
plan() # returns immediately
print(pyfftw.export_wisdom()) # no change
```

First off, the `plan()` function returns way too fast when numpy is imported first. (This is even more obvious if you use the `'FFTW_PATIENT'` flag. If numpy is imported first, the function returns instantly. If numpy is imported second, it takes ~30 minutes, as expected.)

Second, when pyfftw is imported before numpy, the first `pyfftw.export_wisdom()` returns:
```python
(b'(fftw-3.3.4 fftw_wisdom #x3c273403 #x192df114 #x4d08727c #xe98e9b9d\n)\n',
 b'(fftw-3.3.4 fftwf_wisdom #x706526c0 #x2f8b6c85 #x8cd1bb1a #x7c96e03d\n)\n',
 b'(fftw-3.3.4 fftwl_wisdom #x0821b5c7 #xa4c07d5a #x21b58211 #xebe513ab\n)\n')
```

but if numpy is imported first, I get:
```python
(b'',
 b'',
 b'(fftw-3.3.4 fftwl_wisdom #x0821b5c7 #xa4c07d5a #x21b58211 #xebe513ab\n)\n')
```

And after the call to `plan()`, there is no change to the wisdom. (If pyfftw is imported first, calling `plan()` adds a ton of wisdom.)

[...]

Issue #127, opened 2016-06-07
Runtime failure on architectures without long-double FFTW


ghisvail:
I have updated the [Debian packaging](https://buildd.debian.org/status/package.php?p=pyfftw) with release `0.10.4` and the build now fails for some architectures, some of these considered major. On `armel`, `armhf` and `mips[el]`, it seems that importing the module triggers an import of the long-double FFTW although the latter is not available and not being built against.

hgomersall:
Ok - did this work previously? Is it possible to see what change causes it by comparison to earlier releases?

ghisvail:
The testsuite was disabled previously because of #114. Now that the testsuite is enabled as part of the package build, the errors are coming.

ghisvail:
This is the patch I am currently applying to get the build to work for architectures without long-double FFTW:
```
From: Ghislain Antony Vaillant <ghisvail@gmail.com>
Date: Mon, 1 Feb 2016 23:08:52 +0000
Subject: Detect fftw libraries using pkgconfig.

---
 setup.py | 6 ++++--
 1 file changed, 4 insertions(+), 2 deletions(-)

--- a/setup.py
+++ b/setup.py
@@ -120,8 +120,10 @@
         libraries = ['libfftw3-3', 'libfftw3f-3', 'libfftw3l-3']
 
     else:
-        libraries = ['fftw3', 'fftw3f', 'fftw3l', 'fftw3_threads',
-                     'fftw3f_threads', 'fftw3l_threads']
+        import pkgconfig
+        libraries = ['fftw3', 'fftw3_threads', 'fftw3f', 'fftw3f_threads']
+        if pkgconfig.exists('fftw3l'):
+            libraries.extend(['fftw3l', 'fftw3l_threads'])
 
     return libraries

```
But the [outcome](https://buildd.debian.org/status/fetch.php?pkg=pyfftw&arch=armel&ver=0.10.4%2Bdfsg1-1&stamp=1465321508) of running the testsuite for architectures without long-double suggests that this is probably not enough.

hgomersall:
Yes, this is what I suspected. I think this issue of not requiring all the library versions needs to be sorted properly, and is in my long term todo list. Do you agree this bug is essentially the same as #70?

ghisvail:
> Do you agree this bug is essentially the same as #70?

Agreed.

[...]

PR #177, opened 2017-05-15
[setup] Detect available FFTW libs

fredRos:
Test compile to detect libs and headers of the various FFTW data types, threads,
MPI... and build only the pyfftw wrapper for what's available. This requires
cython at build time to avoid missing symbols at runtime.

This is the first milestone needed to integrate the MPI wrapper into master. The tests pass on my local machine. 
@hgomersall Please let me know if this works, I'm open to suggestions here. I noticed you made an effort to hide even the numpy dependency in `setup.py`. Is this because of automated test or doc build environments? At present you make a source release w/o requiring cython but that only works because you require a standard set of FFTW libs. But if users don't have, say the long double versions, then you'd get missing symbols at run time. At least #70 was on that.

I have had users with major problems trying to get the long double version of FFTW built on some super computer just so pyfftw would compile. That's why I implemented the checks in my `EnvironmentSniffer` (we can change that name) to define a dict like 

```python
{'HAVE_DOUBLE_OMP':True, 'HAVE_LONG_THREADS':False}
```
that can be used by cython during build time to figure out what to build. It's similar to how autoconf or cmake define symbols in `config.h`. 

From a user perspective, it's usually very easy to install cython if it isn't available yet. But getting all FFTW libs may be much harder. So I think the cython dependency is not problem.

My contribution also allows to support quad precision in the future. I"m not sure somebody requested that yet but it is doable with fftw already.

I did not use the new information yet to actually determine the building but that is just a cherry-pick away. One would also have to see which unit tests are run. I didn't want to get into that before we have agreement on the changes to setup.py. 

This PR addresses #67, #70


hgomersall:
Yeah, the removal of dependency on libs is to do with not needing them for simply running `setup.py`. The ideal is that `setup.py` should always run irrespective of whether or not the dependencies are installed or not.

It's also worth thinking about how this interacts with #176.

The issues with multiple lib versions goes pretty deep - there is a look up dict that references particular FFTW versions in the cython file. This would obviously be wrong in the case where an fftw version is missing. I think we can use conditional Cython statements based on the introspected variables you've discussed to solve this, but it needs a bit of care to get right.

Thanks!

fredRos:
On 15.05.2017 15:23, Henry Gomersall wrote:
>
> Yeah, the removal of dependency on libs is to do with not needing them
> for simply running |setup.py|. The ideal is that |setup.py| should
> always run irrespective of whether or not the dependencies are
> installed or not.
>
I can make changes to accommodate that. But how do you invoke `setup.py`
to do something useful w/o the dependencies? And in which environment?
>
> It's also worth thinking about how this interacts with #176
> <https://github.com/pyFFTW/pyFFTW/pull/176>.
>
The changes are orthogonal, I didn't touch the version numbers.
>
> The issues with multiple lib versions goes pretty deep - there is a
> look up dict that references particular FFTW versions in the cython
> file. This would obviously be wrong in the case where an fftw version
> is missing. I think we can use conditional Cython statements based on
> the introspected variables you've discussed to solve this, but it
> needs a bit of care to get right.
>
Yes, it requires a lot of care but I have already done that, see for
example here
https://github.com/fredRos/pyFFTW/blob/mpi/pyfftw/pyfftw.pyx#L77
from the mpi branch of my fork. Let's see how many modifications I have
to make for the current version of pyfftw, I branched off more than two
years ago. You may have to help with the new features.

hgomersall:
Ok great re your implementation. It looks exactly right.

There is a circular issue in which setup.py can't run to work out which dependencies are needed in order they can be installed. `setup.py` needs to run without the dependencies so they can be installed.

It might be worth reviewing the changes to `setup.py` - there were some changes to use setuptools, which your patch appears to clobber (specifically, at least `Extension` is overwritten).

You're current version is breaking in python 3 (`iteritems` has been removed - I suspect it's safe to just use `items` which works in both and the list is probably not too painful to use explicitly in 2).

hgomersall:
Build failing :(

fredRos:
I reworked my code to make it cleaner and more robust. The link checks I used to do basically amounted to checking if files were available now I do real links to make sure all libraries are there and they have the right order, which can matter on some systems. You can see the exact commands used in case a test failed in debug mode

    python setup.py -v build_ext -i

The code should now provide the functionality you had before + more. Below I list the open points, let me know if something is wrong or modify the code yourself, I don't plan to do any more structural changes to the code.

> There is a circular issue in which setup.py can't run to work out which dependencies are needed in order they can be installed. setup.py needs to run without the dependencies so they can be installed.

I now hide all import except those in the standard library. But I don't understand how you run `setup.py` to install the dependencies, please let me see the command so we can test that behavior.

> It might be worth reviewing the changes to setup.py - there were some changes to use setuptools, which your patch appears to clobber (specifically, at least Extension is overwritten).

I did, I now use the `Extension` as before. Can' remember why I wanted to use `Cython.Extension`, probably it saved me from calling `cythonize` but that's not important.

> You're current version is breaking in python 3 (iteritems has been removed - I suspect it's safe to just use items which works in both and the list is probably not too painful to use explicitly in 2).

that was trivial to fix. It builds with python 2 and 3 now.

### static linking

Static linking has issues, the tests don't pass, openMP and static linking is a pain, see https://stackoverflow.com/questions/23869981/linking-openmp-statically-with-gcc So where do you want to support static linking? I made quite some effort to include static linking but now I doubt that it's worth it. In the end, you have to create a shared library `pyfftw.so` so we don't have a static include-it-all executable anyways. So why bother with static linking? Removing support for it would simplify the `setup.py`.

### package data

When do you use them? To create a release distribution for windows? I guess in that case it's ok to statically code up the libraries because you have control over the machine where you do the packaging. But is there no threading support on windows? For the moment, I left `get_package_data()` as it is.

### windows

I only tested on debian jessie and ubuntu 16.04. Could you please test the code on windows? I saw that you set up appveyor but on this pr appveyor is not run. I wonder about threads and MPI on windows. It looks like you never supported multiple threads on windows.

### quick test

In my opinion the tests weren't quick, so I commented out all but the first three tests but didn't commit this. Since you use subprocesses, it is not possible to quit all tests once they have been started with `Ctrl-C`, quite annoying.

hgomersall:
This is a really useful bit of work. Thanks! Just to be clear, this basically works out what libraries are available and populates the cython compile time variables? (for a next step commit which makes use of that?)

To answer some of your question:
On windows, posix threads are used. I'm not so worried about Windows as this as the standard means of distribution is a complete package which rolls up the whole lot, so the problem of distributions not including certain dtypes is not there.

The static stuff was added in #116, because of issues with MKL symbols. @grlee77 it would be good to have your opinion on this PR. I suspect we can enforce posix threads in the static case (which still solves the stuff address in #116).

The dependencies stuff has never been dealt with properly, partly because I haven't had time to address it properly before `setup.py` was made runnable. In principle, installations should go and fetch the dependencies for pypi when a `pip install pyfftw` is run. I don't think this currently works, but it should.

The quick tests are definitely quicker than the full tests! Happy to be advised if you think there is a better way to run the tests.

As you have implicitly noted, there is a lot of legacy code that is in need of being improved, much of it from my ignorance of how to do it better. Please feel free to suggest improvements and there is a good chance they will be adopted.

grlee77:
>So where do you want to support static linking?

I added the static linking option as a workaround for the issues reported in #40 and #138.  In short, there can be problems using PyFFTW along with a NumPy that uses the MKL FFT libraries (as in anaconda's default NumPy package).  The [`conda-forge` package for PyFFTW](https://github.com/conda-forge/pyfftw-feedstock) is currently using the static linking option to avoid this problem.

grlee77:
>  I suspect we can enforce posix threads in the static case
I would be fine with that.

grlee77:
> windows
> 
> I only tested on debian jessie and ubuntu 16.04. Could you please test the code on windows? I saw that you set up appveyor but on this pr appveyor is not run.

It is run, it just doesn't get linked into the PR on GitHub.  You can find the Appveyor output here:
https://ci.appveyor.com/project/hgomersall/pyfftw/history
It is currently failing to find the FFTW libraries during setup.

fredRos:
> It is run, it just doesn't get linked into the PR on GitHub. You can find the Appveyor output here:
> https://ci.appveyor.com/project/hgomersall/pyfftw/history
> It is currently failing to find the FFTW libraries during setup.
Great, I turned on debug mode. Will be slow but I can get some info on windows now!

fredRos:
The testing has huge latency when using only appveyor but I'm getting there. There are lots of points, I cover some but not all

> On windows, posix threads are used.

How so? Are the threading routines inside the libs that are shipped? That is, there are no separate libraries for threading as on unix? That would make the detection even more involved... but I have an idea how to do it

> The dependencies stuff has never been dealt with properly, partly because I haven't had time to address it properly before setup.py was made runnable. In principle, installations should go and fetch the dependencies for pypi when a pip install pyfftw is run. I don't think this currently works, but it should.

I just tested it by removing cython from my conda env and ran

    pip install git+https://github.com/fredRos/pyFFTW.git@cond-pr

it installed cython and then pyFFTW just fine.

> I added the static linking option as a workaround for the issues reported in #40 and #138. In short, there can be problems using PyFFTW along with a NumPy that uses the MKL FFT libraries (as in anaconda's default NumPy package). The conda-forge package for PyFFTW is currently using the static linking option to avoid this problem.

I haven't checked that but I didn't run into any issues with the tests in my conda env with numpy MKL. I will seek to choose posix in the static case

fredRos:
Overview
=====

Static linking seems to work. Right now all tests on travis and the python 2 tests on appveyor pass. Please help me with the errors on windows with python 3. They don't seem to be related to python3 itself but rather the build environment

static + threads
------------

I forced the lib to use posix threads with static linking but get this error

```
STATIC_FFTW_DIR=/usr/lib/x86_64-linux-gnu/ CFLAGS="$CFLAGS -Wl,-Bsymbolic" python setup.py -v build_ext -i
...
gcc -pthread -shared -L/home/beaujean/.local/miniconda3/envs/pyfftw3/lib -Wl,-rpath=/home/beaujean/.local/miniconda3/envs/pyfftw3/lib,--no-as-needed build/temp.linux-x86_64-3.6/home/beaujean/workspace/pyfftw/pyfftw/pyfftw.o /usr/lib/x86_64-linux-gnu/libfftw3l_threads.a /usr/lib/x86_64-linux-gnu/libfftw3l.a /usr/lib/x86_64-linux-gnu/libfftw3f_threads.a /usr/lib/x86_64-linux-gnu/libfftw3f.a /usr/lib/x86_64-linux-gnu/libfftw3_threads.a /usr/lib/x86_64-linux-gnu/libfftw3.a -L/home/beaujean/.local/miniconda3/envs/pyfftw3/lib -L/home/beaujean/.local/miniconda3/envs/pyfftw3/lib -lpython3.6m -lm -o build/lib.linux-x86_64-3.6/pyfftw/pyfftw.cpython-36m-x86_64-linux-gnu.so
/usr/bin/ld: /usr/lib/x86_64-linux-gnu/libfftw3l_threads.a(libfftw3l_threads_la-api.o): relocation R_X86_64_32S against `fftwl_mksolver_ct_threads' can not be used when making a shared object; recompile with -fPIC
/usr/lib/x86_64-linux-gnu/libfftw3l_threads.a: error adding symbols: Bad value
collect2: error: ld returned 1 exit status
error: command 'gcc' failed with exit status 1
```

I guess this is expected since I used FFTW from the debian package and they were not compiled with `-fPIC`.

Next I compiled fftw myself with `-fPIC` for single, double, long and threads. Then

```
rm pyfftw/pyfftw.c; STATIC_FFTW_DIR=/home/beaujean/.local/sw/fftw/linux-debian-8.8-x86_64/3.3.6/lib CFLAGS="$CFLAGS -Wl,-Bsymbolic" python setup.py -v build_ext -i
python setup.py -v quick_test
```

work fine. I even tried without setting `Bsymbolic` and that worked, too.

@grlee77 could you please test your conda-forge package with this pr? I don't know how to do that myself.

py 3.4, 32 and 64 bit
-----

``` none
Link FFTW dynamically
objects: []
libraries: []
c:\Program Files (x86)\Microsoft Visual Studio 10.0\VC\BIN\link.exe /nologo /INCREMENTAL:NO /LIBPATH:C:\projects\pyfftw\pyfftw /LIBPATH:C:\Miniconda3\envs\build_env\bin /LIBPATH:C:\Miniconda3\envs\build_env\lib /LIBPATH:C:\Miniconda3\envs\build_env\libs /LIBPATH:C:\Miniconda3\envs\build_env\PCbuild C:\Users\appveyor\AppData\Local\Temp\1\pyfftw-zre62tju\None.obj /OUT:C:\Users\appveyor\AppData\Local\Temp\1\pyfftw-zre62tju\a.out.exe /MANIFESTFILE:C:\Users\appveyor\AppData\Local\Temp\1\pyfftw-zre62tju\a.out.exe.manifest
C:\Program Files (x86)\Microsoft SDKs\Windows\v7.0A\bin\mt.exe -nologo -manifest C:\Users\appveyor\AppData\Local\Temp\1\pyfftw-zre62tju\a.out.exe.manifest -outputresource:C:\Users\appveyor\AppData\Local\Temp\1\pyfftw-zre62tju\a.out.exe;1
C:\Users\appveyor\AppData\Local\Temp\1\pyfftw-zre62tju\a.out.exe.manifest : general error c1010070: Failed to load and parse the manifest. The system cannot find the file specified.
```

No idea why a manifest file is created or needed.

For py 3.5 and 3.6, 32 and 64 bit
---

``` none
C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\BIN\x86_amd64\cl.exe /c /nologo /Ox /W3 /GL /DNDEBUG /MD -IC:\projects\pyfftw\include -IC:\projects\pyfftw\pyfftw -IC:\Miniconda36-x64\envs\build_env\lib\site-packages\numpy\core\include -IC:\Miniconda36-x64\envs\build_env\include -IC:\projects\pyfftw\include\win -IC:\Miniconda36-x64\envs\build_env\include -IC:\Miniconda36-x64\envs\build_env\include /TcC:\projects\pyfftw\pyfftw\pyfftw.c /Fobuild\temp.win-amd64-3.6\Release\projects\pyfftw\pyfftw\pyfftw.obj
pyfftw.c
c:\miniconda36-x64\envs\build_env\include\pyconfig.h(59): fatal error C1083: Cannot open include file: 'io.h': No such file or directory
error: command 'C:\\Program Files (x86)\\Microsoft Visual Studio 14.0\\VC\\BIN\\x86_amd64\\cl.exe' failed with exit status 2
```

I found https://stackoverflow.com/questions/40018405/cython-cannot-open-include-file-io-h-no-such-file-or-directory but I don't know how to apply the fixes. Presumably in `appveyor.yaml`? @hgomersall please help me out.

grlee77:
I am fine with not supporting the static linking option on Windows if it is too much of a pain to get working.  I am not sure if the kind of symbol mixup errors with MKL previously seen on linux even occur on that platform.  The `conda-forge` FFTW and PyFFTW packages only have OS X and linux builds due to the difficulty of building FFTW with MSVC on Windows, so static linking on Windows is not currently a priority as far as conda-forge goes.

If you are intersted in more on the pains of building FFTW on Windows, see discussion in conda-forge/fftw-feedstock#2.  According to the discussion there, FFTW libs will soon be available on Windows through anaconda's own `defaults` channel, but these will be built with msys2/mingw rather than MSVC.  Another comment in that thread indicates that FFTW itself may also support building via CMAKE in the future which would make handling windows builds of the underlying FFTW libraries much easier.

grlee77:
regarding your comment on MKL testing:

> I haven't checked that but I didn't run into any issues with the tests in my conda env with numpy MKL. I will seek to choose posix in the static case

I am not sure any of the existing tests would fail.  What is needed are tests where the number of axes transformed are less than `(ndim - 1)` where `ndim` is the number of array dimensions.

I thought I had made a PR to add such tests, but guess I never submitted it.  Here is a commit that would add a couple cases, but unexpectedly it passed.  I think this may be because I also need to move the numpy import above the pyfftw one at the top of the test so the numpy/MKL symbols get found first.  I will try in a bit and see if I can get one that would fail on the CI services as currently configured when static linking is absent.:
https://github.com/grlee77/pyFFTW/pull/1/commits/4f60c8692d97f5cf93c5fc1255ac92bd8b0a1efa

hgomersall:
All the commands are run in an environment which is set by `appveyor/run_with_env.cmd`. This is configured in the global section of `appveyor.yml` and is executed before each build command.

I concur with @grlee77 - I've seen no bug reports that would suggest we have a problem on windows. For the time being given the difficulty you're having it makes sense to drop it. Windows is much harder to work with that Linux when it comes to any kind of compiling and I feel your pain.

The threading code is built into the dlls. I don't know how it's done (I don't even know how it's compiled). I _think_ it's posix, but it might not be. Basically, I've not had to concern myself so far with this part.

hgomersall:
I guess if we drop this windows bit we just special case windows?

fredRos:
> I am fine with not supporting the static linking option on Windows if it is too much of a pain to get working.

> I concur with @grlee77 - I've seen no bug reports that would suggest we have a problem on windows. For the time being given the difficulty you're having it makes sense to drop it. Windows is much harder to work with that Linux when it comes to any kind of compiling and I feel your pain.

Great. I have no way to test and I'm happy not to have to support it. I could raise an Exception if somebody tried to statically link on windows.

> The threading code is built into the dlls. I don't know how it's done (I don't even know how it's compiled). I think it's posix, but it might not be. Basically, I've not had to concern myself so far with this part.

It is explained here http://fftw.org/install/windows.html and I hope it's up to date. Apparently the release is built with MinGW and one library is created with `./configure --enable-threads --with-combined-threads`.

``` none
./configure --help
...
  --enable-openmp         use OpenMP directives for parallelism
  --enable-threads        compile FFTW SMP threads library
```

So it's not OpenMP on windows! I suppose there are separate libraries for linux only to give users a choice between openmp and posix.

> I thought I had made a PR to add such tests, but guess I never submitted it. Here is a commit that would add a couple cases, but unexpectedly it passed. I think this may be because I also need to move the numpy import above the pyfftw one at the top of the test so the numpy/MKL symbols get found first. I will try in a bit and see if I can get one that would fail on the CI services as currently configured when static linking is absent.:
> grlee77/pyFFTW@4f60c86

Thanks. But I don't insist you run into too much work here. If there is no problem, don't create one!

@hgomersall Any ideas about the py 3.4 and 3.5 errors on appveyor? Have you seen these before? I don't understand why they didn't come up before.

hgomersall:
Ah yes, it's in `BUILD-MINGW64.sh`. Do you mean the manifest errors? I've never seen them before. It seems to be a visual studio issue - the version of visual studio used depends on the python version (up to 3.5 I think).

fredRos:
Error summary
--------------

I'm trying to fix the include errors with python 3.5. Some are missing compared to a build on appveyor for "Merge pull request #173 from grlee77/appveyor_py36_fix"  If I can find out what where these includes come from, I find out what I changed accidentally and hopefully the other errors will vanish, too. The includes are OK in my test compile commands to detect FFTW libraries but some includes are missing when the FFTW extensions is built. I just don't understand why but somehow the compiler (invokation) is changed. I even looked through the distutils source code but couldn't figure out why. In any case, I don't see when the include flags are added, presumably through the environment variables that are set in `appveyor/run_with_env.cmd`. 

Please help me, I'm getting frustrated.

Output
--------

So here is the compile command for pyfftw master branch with python 3.5

```
C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\BIN\cl.exe /c /nologo /Ox /W3 /GL /DNDEBUG /MD -IC:\projects\pyfftw\include -IC:\projects\pyfftw\pyfftw -IC:\Miniconda35\envs\build_env\lib\site-packages\numpy\core\include -IC:\Miniconda35\envs\build_env\include -IC:\projects\pyfftw\include\win -IC:\Miniconda35\envs\build_env\include -IC:\Miniconda35\envs\build_env\include "-IC:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\INCLUDE" "-IC:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\ATLMFC\INCLUDE" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.14393.0\ucrt" "-IC:\Program Files (x86)\Windows Kits\NETFXSDK\4.6.1\include\um" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.14393.0\shared" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.14393.0\um" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.14393.0\winrt" /TcC:\projects\pyfftw\pyfftw\pyfftw.c /Fobuild\temp.win32-3.5\Release\projects\pyfftw\pyfftw\pyfftw.obj

```

It takes some scrolling but all includes in quotes are missing in my branch. These includes are the ones that contain explicit windows kits version strings, for example

    "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.14393.0\winrt" 

When I do compiler checks in my `setup.py`,  the includes are ok. For example, I call

```python
        log.debug("objects: %s" % objects)
        log.debug("libraries: %s" % libraries)
        log.debug("include dirs: %s" % include_dirs)
...
        self.compiler.compile([fname], output_dir=file_root, include_dirs=include_dirs)
```

and get

```
objects: []
libraries: ['libfftw3q-3', 'libfftw3l-3', 'libfftw3f-3', 'libfftw3-3']
include dirs: ['C:\\projects\\pyfftw\\include', 'C:\\projects\\pyfftw\\pyfftw', 'C:\\Miniconda35-x64\\envs\\build_env\\lib\\site-packages\\numpy\\core\\include', 'C:\\Miniconda35-x64\\envs\\build_env\\include', 'C:\\projects\\pyfftw\\include\\win']
C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\BIN\x86_amd64\cl.exe /c /nologo /Ox /W3 /GL /DNDEBUG /MD -IC:\projects\pyfftw\include -IC:\projects\pyfftw\pyfftw -IC:\Miniconda35-x64\envs\build_env\lib\site-packages\numpy\core\include -IC:\Miniconda35-x64\envs\build_env\include -IC:\projects\pyfftw\include\win -IC:\Miniconda35-x64\envs\build_env\include -IC:\Miniconda35-x64\envs\build_env\include "-IC:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\INCLUDE" "-IC:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\ATLMFC\INCLUDE" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.14393.0\ucrt" "-IC:\Program Files (x86)\Windows Kits\NETFXSDK\4.6.1\include\um" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.14393.0\shared" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.14393.0\um" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.14393.0\winrt" /TcC:\Users\appveyor\AppData\Local\Temp\1\pyfftw-joda4wwd\fftwq_plan_dft.c /FoC:\Users\appveyor\AppData\Local\Temp\1\pyfftw-joda4wwd\fftwq_plan_dft.obj
fftwq_plan_dft.c
C:\Users\appveyor\AppData\Local\Temp\1\pyfftw-joda4wwd\fftwq_plan_dft.c(2): warning C4013: 'fftwq_plan_dft' undefined; assuming extern returning int
```

but when the `pyfftw` extension is compiled, there are only the includes that I added myself

```
creating build\temp.win-amd64-3.5\Release\projects\pyfftw\pyfftw
C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\BIN\x86_amd64\cl.exe /c /nologo /Ox /W3 /GL /DNDEBUG /MD -IC:\projects\pyfftw\include -IC:\projects\pyfftw\pyfftw -IC:\Miniconda35-x64\envs\build_env\lib\site-packages\numpy\core\include -IC:\Miniconda35-x64\envs\build_env\include -IC:\projects\pyfftw\include\win -IC:\Miniconda35-x64\envs\build_env\include -IC:\Miniconda35-x64\envs\build_env\include /TcC:\projects\pyfftw\pyfftw\pyfftw.c /Fobuild\temp.win-amd64-3.5\Release\projects\pyfftw\pyfftw\pyfftw.obj
pyfftw.c
c:\miniconda35-x64\envs\build_env\include\pyconfig.h(68): fatal error C1083: Cannot open include file: 'io.h': No such file or directory
error: command 'C:\\Program Files (x86)\\Microsoft Visual Studio 14.0\\VC\\BIN\\x86_amd64\\cl.exe' failed with exit status 2
Command exited with code 1
```

fredRos:
No clues from anyone? I ran `python setup.py build_ext -i` through a debugger on linux. A vague guess: in my own check, I call `self.compiler.compile` on a source file but perhaps on windows a different instance of `ccompiler` is created when building a shared library? I saw that there also is a  `shlib_compiler` in in `setuptools/.../build_ext.py`. Perhaps that is treated differently on windows??

hgomersall:
Hey, Nice. Appveyor is building this now. Does that suggest the problem is solved?

fredRos:
Let's wait to see if all builds pass. I missed some include directories that are automatically set for py 3.5 and 3.6 and even after staring at the distutils code I still don't understand how they are set but now we use them. The fix for 3.4 is different and to me seems independent of my changes, I"m surprised that previous tests pass. It seems to be this old but unfixed bug http://bugs.python.org/issue4431

I now have many tiny commits just for appveyor debugging. If you are happy,  I recommend you squash all commits into a single commit for this PR

fredRos:
From my side, I'm done. Please approve the review and merge

setup.py
```diff
@@ -1,8 +1,11 @@
-# Copyright 2015 Knowledge Economy Developments Ltd
+# Copyright 2017 Henry Gomersall and Frederik Beaujean
```
grlee77: Should this maybe be something generic like "The PyFFTW Contributors"?  I know a few others have made contributions to this file in the past as well.

setup.py
```diff
@@ -32,56 +35,39 @@
...
+from distutils import log
+from distutils.ccompiler import get_default_compiler, new_compiler
+from distutils.errors import CompileError, LinkError
+from distutils.extension import Extension
+from distutils.sysconfig import customize_compiler
+from distutils.util import get_platform
 
-import os
-import sys
+import contextlib, os, sys
```
grlee77: PEP8 suggests these be on 3 separate lines

setup.py
```diff
@@ -32,56 +35,39 @@
...
-#    git_tag = os.environ.get('TRAVIS_TAG')
-#elif os.environ.get('APPVEYOR_REPO_TAG_NAME') is not None:
-#    git_tag = os.environ.get('APPVEYOR_REPO_TAG_NAME')
-#else:
-#    git_tag = None
+# we require cython because we need to know which part of wrapper to build to
+# avoid missing symbols at run time. But if this script is called without
+# building pyfftw, then we may hide cython dependency.
+# TODO Drop zig-zag to avoid cython dependency below
+# from Cython.Distutils import build_ext
```
grlee77: remove outdated comment?  What is "zig-zag" referring to?

setup.py
```diff
@@ -121,6 +95,20 @@ def get_include_dirs():
 
     return include_dirs
 
+# TODO Do we need to determine package data dynamically? If so, should
```
grlee77: ideally we would determine dynamically, but I think that could be left to a future PR if it ends up being needed.  One example is the proposed CMAKE-based Windows FFTW build in conda-forge/fftw-feedstock#32 which does not build the long-double variant and the library names end up being just `fftw3.dll` and `fftw3f.dll`.

setup.py
fredRos: Ok with me. I don't insist on having my name in there. Just wanted to make sure people can find me if they have questions about what and why. Perhaps there should be a list of contributors at the end of the docs that list the PyFFTW contributors?

setup.py
fredRos: Ok, I fix it

setup.py
fredRos: I removed the zig-zag and moved the comment close to the `import cython` line

setup.py
fredRos: I agree to to leave it as it is. You are responsible for the conda pkg, right? So now you are aware and the TODO will help you remember if you or someone else creates the PR

fredRos:
> Why link to the library if it isn't going to be used?

Good catch, you are right. This is a relict from my branch where MPI wrappers are implemented. I set it it to False and added a `TODO` to enable it in the next round.

setup.py
grlee77: ok.  yes, I am one of the maintainers on the conda package.

setup.py
grlee77: Certainly you will receive credit.  The question is just the best way to handle it.  Let's see what @hgomersall thinks.  I don't think it makes sense to add all names to individual files, as this can easily get out of date, thus something like "PyFFTW Contributors" or "Henry Gomersall and the PyFFTW contributors" (to reflect his much larger overall contribution to date) seems reasonable.

If someone accesses the file via Github, it shows who has previously edited it, so I think that will be reasonably clear either way.  (However, oddly, when checking https://github.com/pyFFTW/pyFFTW/blob/master/setup.py just now, I don't see my username there despite changes to this file in #116 for the static linking).  Obviously, you will also receive credit via the autogenerated listings at https://github.com/pyFFTW/pyFFTW/graphs/contributors.

grlee77:
To fully address #70 and #127 it seems we would also need to build in some way of detecting at runtime which libraries are available.  For instance, the tests currently run using float, double and long double dtypes so they will fail if pyfftw was built without any of these.

Fixing that can be the subject of a separate PR, but do you have an idea of how one might best to approach that?

fredRos:
> Fixing that can be the subject of a separate PR,

Definitely. That's how it was planned from the start.

> do you have an idea of how one might best to approach that?

I'm glad to discuss. Here's how I did in the mpi branch, see https://github.com/fredRos/pyFFTW/blob/mpi/test/test_pyfftw_mpi.py. I hadn't touched the other tests, we should agree on the pattern first because it involves a lot of tedious editing

```python
try:
    from pyfftw import FFTW_MPI, broadcast_wisdom, create_mpi_plan, \
                       gather_wisdom, local_size, \
                       supported_mpi_types

    mpi_import_failed = False
except ImportError:
    mpi_import_failed = True
...
    def test_threads(self):
        if '32' not in supported_mpi_types:
            self.skipTest('single precision not build')
```

I supposed it would be cleaner to use decorators around the test.

But perhaps I should bring the MPI tests into the same form as the other tests. From a quick glance, I guess a skip somewhere in `FFTWBaseTest` would do the trick. In any case, the available types have to be available for introspection at runtime.

grlee77:
Hey @fredRos .  I made a small PR to your branch with fixes to a couple of issues I came across when testing the current [pyfftw-feedstock recipe](https://github.com/conda-forge/pyfftw-feedstock) on `conda-forge` using this branch.

@hgomersall, I think this PR is a great addition and is pretty much ready to go.  There may be some other corner cases we have not anticipated, but probably the best way to find those is to merge this and then fix them as they crop up.

hgomersall:
Great, thanks all.

grlee77:
That was fast!  My recent commits proposed to @fredRos branch had not yet been merged/reviewed.  I will reopen a PR for them over here instead.

setup.py
```diff
@@ -464,6 +772,15 @@ def setup_package():
...
         build_requires = ['numpy>=1.6, <2.0']
 
+    # we require cython because we need to know which part of the wrapper
+    # to build to avoid missing symbols at runtime. But if this script is
+    # called without building pyfftw, for example to install the
+    # dependencies, then we have to hide the cython dependency.
+    try:
+        import cython
+    except ImportError:
+        build_requires.append('cython>=0.23, <1.0')
```
jakirkham: Since we are needing Cython anyways, we might consider hardening the requirement on Setuptools and dropping C source from the codebase. The reason being Setuptools (as of [version 18.0]( https://github.com/pypa/setuptools/blob/master/CHANGES.rst#180 ), quite old now) supports compiling the Cython code for us as long as we add Cython to `setup_requires` and point out the Cython source code. Expect that doing this will vastly simplify what we are doing in `setup.py` and simplify our release process (as we are approaching that stage).

hgomersall: haha, this has gone round in circles. I think it was @ghisvail who originally advocated including the C source. I honestly don't really care about this (including or not).

jakirkham: Interesting. Would be curious to know more about the argument for it (and whether it still applies).

Normally would think that packagers would prefer to be as close to the source of truth as possible. In our case, that is the Cython code. There is always the possibility we forget to generate updated C code. However the same can't be said of the Cython code, which must be up to date and functioning.

grlee77: > Interesting. Would be curious to know more about the argument for it (and whether it still applies).

see pyFFTW/pyFFTW#31 and PyWavelets/pywt#321 for the prior discussion.

grlee77: Also from one of those prior conversations I had found this link to the Cython docs where they recommend distributing the C files:
http://cython.readthedocs.io/en/latest/src/reference/compilation.html#distributing-cython-modules

jakirkham: Thanks for the references.

My reading of that issue is that @ghisvail and I would agree.

Have heard people argue for and against the inclusion of C code matching Cython. From my perspective, it's best to keep only one representation of the source around. Since we have made some improvements on the dependency declaration end, expect we will resolve the user issues that prompted inclusion of C code. That said, user problems likely would go away if we just provided binaries.

At the end of the day, don't have particularly strong feelings about this.

fredRos: Since my pr #189 has been merged, we need cython to build because we are now flexible to build if say long double precision is missing. So we shouldn't distribute C source files in the sdist.

hgomersall: My bad, I got @ghisvail's position wrong (he's agreeing with you!). I think our current position is reasonable a low cost, as pointed out by @rgommers in https://github.com/PyWavelets/pywt/pull/321.

hgomersall: Ach, I just read the whole thread. I can definitely see @fredRos's point. What happens currently if we include it and long double is missing?

fredRos: If long double is missing, different C code is generated from cython by making using of preprocessor-style variables defined in `setup.py`. Here is an example from `pyfftw.pyx`. Note the capital `IF`

```cython
IF HAVE_LONG:
    _supported_types.append('ld')
    _supported_nptypes_complex.append(np.clongdouble)
    _supported_nptypes_real.append(np.longdouble)
```
So you see that we definitely need cython to build

hgomersall: Ok, I think I'm convinced as to why we should not include the C file and depend on Cython. There is a clear requirement for explicitly building from the cython source in every case.

PR #178, opened 2017-05-17
Disable norm as a keyword argument in Dask test

jakirkham:
Add another `has_norm_kwarg` to another test of the Dask FFT interface as it the Dask functions do not accept the `norm` keyword argument.


hgomersall:
Fine!

PR #182, opened 2017-07-10
 include documentation in the source distribution

### LINKED ISSUE #31: Cython generated c files should not be part of the source distribution
From  a development standpoint, any auto-generated files should not be generated as they can be refreshed at build time. Cython should be made a build dependency in the `setup.py` file and c-files should be excluded from `MANIFEST.in`.

From a packaging standpoint, the build system will be called twice, one for the  Python 2 build and another one for Python 3. Each time, the c-files will be refreshed, overwriting the ones currently provided by the source distribution. This generates unnecessary pain for packaging.

### LINKED ISSUE #32: Provide the sources of the documentation in the source distribution instead of a build
Similar to #31 but for the documentation. Again, please let the build system build the documentation. Sphinx is still evolving and produce different files as a result from one system to another.

For Debian in particular, some of the javascript code may need to be patched for security reasons if the documentation is produced using an old version of sphinx. As a result, it is advised to provide the source instead of a build and let the packaging system handle the build and potential patching.

In terms of changes, it should be just a matter of modifying the `MANIFEST.in` file and regenerate + upload the source distribution with `python setup.py register sdist upload`.

### LINKED ISSUE #86: Add the source of the documentation to the release tarball.
Please consider adding the source used for generating the HTML documentation in your release tarball, so that a documentation package can be built by Linux distributions. The pre-built HTML documentation you are currently shipping cannot be used directly for reproducibility reasons.

Cheers,
Ghis

grlee77:
Updates MANIFEST.in so the docs will get properly included in the sdist.

~~While I was there, I also addressed #31 (exclude cython-generated files from the source distribution)~~

closes #32
closes #86

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -17,9 +17,6 @@ recursive-include docs *
 # Exclude built docs
 prune docs/_build
 
-# exclude Cython-generated source files
-exclude pyfftw/pyfftw.c
-
 # exclude built libraries, swap files, etc.
 prune build
 prune */__pycache__
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -9,6 +9,19 @@ include pyfftw/utils.pxi
 include test/*.py
 include conf.py
 include index.rst
-recursive-include pyfftw *.rst
-recursive-include sphinx *.rst
 recursive-include include *.h
+
+# All documentation
+recursive-include docs *
+
+# Exclude built docs
+prune docs/_build
+
+# exclude Cython-generated source files
+exclude pyfftw/pyfftw.c
+
+# exclude built libraries, swap files, etc.
+prune build
+prune */__pycache__
+global-exclude *.py[cod] *.so *.dll *.dylib
+global-exclude *~ *.bak *.swp


hgomersall:
I think this is fine. Can you envisage any way it can break anything?

grlee77:
I don't think is should break anything, but Cython will be required when building from source.  Cython is already listed in the requirements.txt and is easily obtained so I don't think that is a problem.  

I tested locally that I was able to build from source using the generated sdist on both linux and OS X, so I am pretty sure nothing critical is missing.

grlee77:
I am rethinking whether it is really a good idea to exclude the generated pyfftw.c from the source distribution as suggested in #31.  

See also the related discussion in https://github.com/PyWavelets/pywt/pull/321.  Going forward we will want to also add a similar `pyproject.toml` file to PyFFTW for PEP 518 support (to be included in pip 10).

grlee77:
Based on the recommendation in the Cython docs and the discussion in PyWavelets/pywt#321, I restored the prior behavior of including the Cython-generated pyfftw.c in the sdist.  Now this PR only deals with including the documentation.

hgomersall:
Thanks for that. I have no strong opinion either way on the cython code - I'm sure there was some rationale at the beginning for what it is now but if new evidence comes to light we can change it.

Apologies for my delay on stuff, I've been a bit swamped with various work pressures as well as trying to prepare our house for moving (and then moving!).

PR #186, opened 2017-08-24
update interfaces.scipy_fftpack namespace

grlee77:
update the `scipy` interface to reflect the current `scipy` minimum version.  For 0.12 as specified in `requirements.txt`, `dst` and `idst` will always be present.  Some other functions such as `fftshift`, `ifftshift` were also missing from `__all__` and have been added.

In the second commit, I add a check for four new n-dimensional routines that are present in current `scipy `master and will be released with `scipy` 1.0

### DIFF
--- a/pyfftw/interfaces/scipy_fftpack.py
+++ b/pyfftw/interfaces/scipy_fftpack.py
@@ -60,7 +60,7 @@ from . import numpy_fft
 import numpy
 
 # Complete the namespace (these are not actually used in this module)
-from scipy.fftpack import (dct, idct, diff, tilbert, itilbert,
+from scipy.fftpack import (dct, idct, dst, idst, diff, tilbert, itilbert,
         hilbert, ihilbert, cs_diff, sc_diff, ss_diff, cc_diff,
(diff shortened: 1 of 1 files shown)


grlee77:
Most likely, we can also close gh-59 which originally opened to address problems when `dct` or `idct` are not present, but these have existed since `scipy` 0.8 or so and our minimum `scipy` requirement is now specified as 0.11.0

hgomersall:
Hmm, appveyor failure...

grlee77:
Can you try just restarting the one failed case on appveyor?  It looks like a download failure occured in conda before the build even started.  I tried to restart it, but don't appear to have permissions for pyFFTW on Appveyor.  I am not sure where that is configured, but I think you can grant Administrators or Restarter permissions to individual users somewhere in the settings.

grlee77:
Is there a good reason to keep Appveyor at:
https://ci.appveyor.com/project/hgomersall/pyfftw
rather than
https://ci.appveyor.com/project/pyfftw/pyfftw
I'm not sure if the latter would automatically grant the permission to all pyFFTW team members.  Also not sure how much of a pain it would be to switch, but something to consider.

Issue #27, opened 2013-10-25
missing float and long-double libraries


aberndsen:
Similar to the issues in
https://github.com/hgomersall/pyFFTW/issues/16

It would be nice to remove the dependencies on the long-double fftw3l (or single fftw3f) libraries. Or, at least provide a description on how to get pyfftw to work if these libraries are missing (not compiled). I suppose the current solution is to compile fftw3 with both the single and long-double options. Correct?

jakirkham:
This was solved with PR ( https://github.com/pyFFTW/pyFFTW/pull/189 ), correct @fredRos? If so, think we should close.

fredRos:
Yes, just about 4 years later this issue is solved

jakirkham:
Thanks. :)

PR #189, opened 2017-08-29
[setup, pyx] Build only the parts for which FFTW libraries were found

### LINKED ISSUE #127: Runtime failure on architectures without long-double FFTW

fredRos:
After #177 and #185 have been merged, it remains to use available information on available libs during build time and during run time to run only available tests. 

This is work in progress and should not be merged yet but I want to open the PR so you can have a look at what's coming.

The essential steps

- [x] build with cython compile time `IF HAVE_DOUBLE`. Fixes long-standing issues #27 and #70
- [x] skip tests for unavailable FFTW libs. Fixes #127

### DIFF
--- a/pyfftw/builders/_utils.py
+++ b/pyfftw/builders/_utils.py
@@ -52,6 +52,7 @@ and may change in future versions.
 
 import pyfftw
 import numpy
+import warnings
 
 __all__ = ['_FFTWWrapper', '_rc_dtype_pairs', '_default_dtype', '_Xfftn',
         '_setup_input_slicers', '_compute_array_shapes', '_precook_1d_args',
@@ -67,7 +68,7 @@ _valid_efforts = ('FFTW_ESTIMATE', 'FFTW_MEASURE',
 _rc_dtype_pairs = {}
 _default_dtype = None
 
-# Double precision is the default. Prefer casting to higher precision if
+# Double precision is the default default precision. Prefer casting to higher precision if
 # possible. If missing, long double is mapped to double and so we lose less
 # precision than by converting to single.
 if '64' in pyfftw._supported_types:
@@ -134,18 +135,23 @@ def _Xfftn(a, s, axes, overwrite_input,
 
     a_is_complex = numpy.iscomplexobj(a)
 
-    # Make the input dtype correct
+    # Make the input dtype correct by transforming to an available type
     if a.dtype.char not in _rc_dtype_pairs:
-        dtype_char = _default_dtype.char
+        dtype = _default_dtype
         if a.dtype == numpy.dtype('float16') and '32' in pyfftw._supported_types:
             # convert half-precision to single precision, if available
-            dtype_char = numpy.dtype('float32').char
+            dtype = numpy.dtype('float32')
+
+        # warn when losing precision but not when using a higher precision
+        if dtype.itemsize < a.dtype.itemsize:
+            warnings.warn("Narrowing conversion from %s to %s precision" % (a.dtype, dtype))
+
         if not real or inverse:
             # It's going to be complex
-            dtype_char = _rc_dtype_pairs[dtype_char].char
+            dtype = numpy.dtype(_rc_dtype_pairs[dtype.char])
 
-        # convert the input array
-        a = numpy.asarray(a, dtype=dtype_char)
+        # finally convert the input array
+        a = numpy.asarray(a, dtype=dtype)
     elif not (real and not inverse) and not a_is_complex:
         # We need to make it a complex dtype
         a = numpy.asarray(a, dtype=_rc_dtype_pairs[a.dtype.char])
--- a/pyfftw/builders/_utils.py
+++ b/pyfftw/builders/_utils.py
@@ -46,7 +46,7 @@ the scenes.
 Certainly, users may encounter instances of
 :class:`~pyfftw.builders._utils._FFTWWrapper`.
 
-These everything documented in this module is *not* part of the public API
+Everything documented in this module is *not* part of the public API
 and may change in future versions.
 '''
 
@@ -60,18 +60,35 @@ __all__ = ['_FFTWWrapper', '_rc_dtype_pairs', '_default_dtype', '_Xfftn',
 _valid_efforts = ('FFTW_ESTIMATE', 'FFTW_MEASURE',
         'FFTW_PATIENT', 'FFTW_EXHAUSTIVE')
 
-# Looking up a dtype in here returns the complex complement of the same
-# precision.
+# Looking up a real dtype in here returns the complex complement of the same
+# precision, and vice versa.
 # It is necessary to use .char as the keys due to MSVC mapping long
 # double to double and the way that numpy handles this.
-_rc_dtype_pairs = {numpy.dtype('float32').char: numpy.dtype('complex64'),
-        numpy.dtype('float64').char: numpy.dtype('complex128'),
-        numpy.dtype('longdouble').char: numpy.dtype('clongdouble'),
-        numpy.dtype('complex64').char: numpy.dtype('float32'),
-        numpy.dtype('complex128').char: numpy.dtype('float64'),
-        numpy.dtype('clongdouble').char: numpy.dtype('longdouble')}
+_rc_dtype_pairs = {}
+_default_dtype = None
 
-_default_dtype = numpy.dtype('float64')
+# Double precision is the default. Prefer casting to higher precision if
+# possible. If missing, long double is mapped to double and so we lose less
+# precision than by converting to single.
+if '64' in pyfftw._supported_types:
+    _default_dtype = numpy.dtype('float64')
+    _rc_dtype_pairs.update({
+        numpy.dtype('float64').char: numpy.dtype('complex128'),
+        numpy.dtype('complex128').char: numpy.dtype('float64')})
+if 'ld' in pyfftw._supported_types:
+    if _default_dtype is None:
+        _default_dtype = numpy.dtype('longdouble')
+    _rc_dtype_pairs.update({
+         numpy.dtype('longdouble').char: numpy.dtype('clongdouble'),
+         numpy.dtype('clongdouble').char: numpy.dtype('longdouble')})
+if '32' in pyfftw._supported_types:
+    if _default_dtype is None:
+        _default_dtype = numpy.dtype('float32')
+    _rc_dtype_pairs.update({
+        numpy.dtype('float32').char: numpy.dtype('complex64'),
+        numpy.dtype('complex64').char: numpy.dtype('float32')})
+if _default_dtype is None:
+    raise NotImplementedError("No default precision available")
 
 
 def _unitary(norm):
@@ -119,22 +136,16 @@ def _Xfftn(a, s, axes, overwrite_input,
 
     # Make the input dtype correct
     if a.dtype.char not in _rc_dtype_pairs:
-        if a.dtype == numpy.dtype('float16'):
-            # convert half-precision to single precision rather than double
-            if not real or inverse:
-                a = numpy.asarray(
-                    a, dtype=_rc_dtype_pairs[numpy.dtype('float32').char])
-            else:
-                a = numpy.asarray(a, dtype=numpy.dtype('float32').char)
-        else:
-            # We make it the default dtype
-            if not real or inverse:
-                # It's going to be complex
-                a = numpy.asarray(
-                    a, dtype=_rc_dtype_pairs[_default_dtype.char])
-            else:
-                a = numpy.asarray(a, dtype=_default_dtype)
+        dtype_char = _default_dtype.char
+        if a.dtype == numpy.dtype('float16') and '32' in pyfftw._supported_types:
+            # convert half-precision to single precision, if available
+            dtype_char = numpy.dtype('float32').char
+        if not real or inverse:
+            # It's going to be complex
+            dtype_char = _rc_dtype_pairs[dtype_char].char
 
+        # convert the input array
+        a = numpy.asarray(a, dtype=dtype_char)
     elif not (real and not inverse) and not a_is_complex:
         # We need to make it a complex dtype
         a = numpy.asarray(a, dtype=_rc_dtype_pairs[a.dtype.char])
--- a/pyfftw/interfaces/numpy_fft.py
+++ b/pyfftw/interfaces/numpy_fft.py
@@ -39,13 +39,15 @@ of :mod:`numpy.fft`, but those functions that are not included here are imported
 directly from :mod:`numpy.fft`.
 
 
-It is notable that unlike :mod:`numpy.fftpack`, these functions will
-generally return an output array with the same precision as the input
-array, and the transform that is chosen is chosen based on the precision
-of the input array. That is, if the input array is 32-bit floating point,
-then the transform will be 32-bit floating point and so will the returned
-array. Half precision input will be converted to single precision.  Otherwise,
-if any type conversion is required, the default will be double precision.
+It is notable that unlike :mod:`numpy.fftpack`, these functions will generally
+return an output array with the same precision as the input array, and the
+transform that is chosen is chosen based on the precision of the input array.
+That is, if the input array is 32-bit floating point, then the transform will
+be 32-bit floating point and so will the returned array. Half precision input
+will be converted to single precision. Otherwise, if any type conversion is
+required, the default will be double precision. If pyFFTW was not built with
+support for double precision, the default is long double precision. If that is not
+available, it defaults to single precision.
 
 One known caveat is that repeated axes are handled differently to
 :mod:`numpy.fft`; axes that are repeated in the axes argument are considered
--- a/test/test_pyfftw_numpy_interface.py
+++ b/test/test_pyfftw_numpy_interface.py
@@ -229,6 +229,16 @@ class InterfacesNumpyFFTTestFFT(unittest.TestCase):
 
         np_input_array = numpy.asarray(input_array)
 
+        # Why are long double inputs copied to double precision? It's what
+        # numpy silently does anyways as of v1.10 but helps with backward
+        # compatibility and scipy.
+        # https://github.com/pyFFTW/pyFFTW/pull/189#issuecomment-356449731
+        if np_input_array.dtype == 'clongdouble':
+            np_input_array = numpy.complex128(input_array)
+
+        elif np_input_array.dtype == 'longdouble':
+            np_input_array = numpy.float64(input_array)
+
         with warnings.catch_warnings(record=True) as w:
             # We catch the warnings so as to pick up on when
             # a complex array is turned into a real array
@@ -265,7 +275,7 @@ class InterfacesNumpyFFTTestFFT(unittest.TestCase):
                 return
             try:
                 output_array = getattr(self.test_interface, self.func)(
-                                    copy_func(input_array), s, **kwargs)
+                                    copy_func(np_input_array), s, **kwargs)
             except NotImplementedError as e:
                 # check if exception due to missing precision
                 msg = repr(e)
@@ -285,23 +295,10 @@ class InterfacesNumpyFFTTestFFT(unittest.TestCase):
                 numpy.allclose(output_array, test_out_array,
                     rtol=1e-2, atol=1e-4))
 
-        # print(numpy.asanyarray(input_array).real.dtype, "")
-        # print(_all_types_np)
-        if _all_types_np.get(numpy.asanyarray(input_array).real.dtype, "") in _supported_types:
+        if _all_types_np.get(np_input_array.real.dtype, "") in _supported_types:
             # supported precisions should not be converted
             self.assertEqual(np_input_array.real.dtype,
                              output_array.real.dtype)
-        # else:
-        #     # a warning will be issued
-        #     with warnings.catch_warnings(record=True) as w:
-        #         # Cause all warnings to always be triggered.
-        #         warnings.simplefilter("always")
-        #         print(np_input_array.real.dtype)
-        #         input_precision_dtype = numpy.asanyarray(input_array).real.dtype
-        #         self.assertEqual(len(w), 1)
-        #         self.assertTrue(issubclass(w[0].category, UserWarning))
-        #         self.assertTrue("Converting input" in str(w[0].message))
-
 
         if (not self.overwrite_input_flag in kwargs or
                 not kwargs[self.overwrite_input_flag]):
--- a/test/test_pyfftw_numpy_interface.py
+++ b/test/test_pyfftw_numpy_interface.py
@@ -32,7 +32,7 @@
 # POSSIBILITY OF SUCH DAMAGE.
 #
 
-from pyfftw import interfaces, _supported_types
+from pyfftw import interfaces, _supported_types, _all_types_np
 
 from .test_pyfftw_base import run_test_suites
 from ._get_default_args import get_default_args
@@ -229,11 +229,12 @@ class InterfacesNumpyFFTTestFFT(unittest.TestCase):
 
         np_input_array = numpy.asarray(input_array)
 
-        if np_input_array.dtype == 'clongdouble':
-            np_input_array = numpy.complex128(input_array)
+        # TODO why are long double inputs copied to double precision?
+        # if np_input_array.dtype == 'clongdouble':
+        #     np_input_array = numpy.complex128(input_array)
 
-        elif np_input_array.dtype == 'longdouble':
-            np_input_array = numpy.float64(input_array)
+        # elif np_input_array.dtype == 'longdouble':
+        #     np_input_array = numpy.float64(input_array)
 
 
         with warnings.catch_warnings(record=True) as w:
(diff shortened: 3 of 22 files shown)


grlee77:
Thanks @fredRos!  This looks pretty good to me.

Based on this PR, I was able to build pyFFTW using the windows conda-forge package which currently does not contain the long double variant.

This also required a couple of extra edits to `setup.py` for compatibility conda-forge `fftw` Windows library names/locations, but that is a separate topic from this PR.  I will make a separate small PR for conda-forge Windows `fftw` compatibility.

test/test_pyfftw_numpy_interface.py
```diff
@@ -50,8 +50,20 @@
...
+    real_dtypes.extend([numpy.float16, numpy.float32])
+if '64' in _supported_types:
+    complex_dtypes.append(numpy.complex128)
+    real_dtypes.append(numpy.float64)
+if 'ld' in _supported_types:
+    complex_dtypes.append(numpy.clongdouble)
+    real_dtypes.append(numpy.longdouble)
+
+# complex_dtypes = (numpy.complex64, numpy.complex64, numpy.complex128, numpy.clongdouble)
+# real_dtypes = (numpy.float16, numpy.float32, numpy.float64, numpy.longdouble)
```
grlee77: Nice.
Unless I missed it, you will also need to fix this in the scipy interfaces for versions < 0.19 because those currently don't import from these numpy ones.:

https://github.com/pyFFTW/pyFFTW/blob/60b81051e4a09a7fd599abd3123aaaa6fe02932d/test/test_pyfftw_scipy_interface.py#L82

You could import from numpy ones there and just drop the unsupported `float16', 'complex64' pair.  (float16 support in scipy.fftpack was fixed in 0.19)

fredRos: Yes, this is still work in progress. If all libraries are available as before, all tests pass. But if I build w/o single precision, there are still multiple tests I need to patch, scipy is one of them. I have to think of a good way of automated testing in that case. I may have to add environment variables that `setup.py` picks up to ignore for example single precision.

fredRos:
All tests pass. On top of the usual test suite, I run tests on travis to build leaving out one precision at a time. I don't do this on appveyor because there we assume the official windows build of FFTW that comes with all three precisions.

Ready for merge!

hgomersall:
This is great work and seems to be the correct way to do it. The issue I currently see is the builders don't seem to be updated to reflect the available types. Specifically, the `_rc_dtype_pairs` in [builders._utils](https://github.com/pyFFTW/pyFFTW/blob/master/pyfftw/builders/_utils.py) should reflect the available dtypes. That dict is used [here](https://github.com/pyFFTW/pyFFTW/blob/master/pyfftw/builders/_utils.py#L121) and we need to confirm that logic is still correct in light of a dynamic set of dtypes (it might well be).

Correct me if I'm wrong on all that!

grlee77:
>  That dict is used here and we need to confirm that logic is still correct in light of a dynamic set of dtypes (it might well be).

In that file, `_default_dtype` is [set to double precision](https://github.com/pyFFTW/pyFFTW/blob/master/pyfftw/builders/_utils.py#L74), but that would be a problem if double precision isn't available.  Also, float16 gets converted to float32, but needs to have a fallback to `_default_dtype` in cases where float32 isn't available.

When setting `_default_dtype`, what would be your prefered order of priority?  When float64 isn't available do we want to prioritize precision (long double) or speed/memory (single)?

fredRos:
In my opinion, there is nothing to change, all tests involving the builders either pass or are skipped. I added another test that ensures pyFFTW fails with an informative error message if a precision is not available. I decided against casting or converting to a different precision. That would add a whole lot of extra work that I find unnecessary.

grlee77:
Okay, that sounds reasonable. The error message should make it clear that recompilation with the desired precision is necessary.  When using the builders and trying to get maximum performance you probably don't want things converting to different precision automatically anyways as that is going to result in performance loss.

fredRos:
The error message now is

    "The scheme '('c2c', 'ld')' is not supported.
    Rebuild pyFFTW with support for long double precision!"

I guess we should update the installation instructions

hgomersall:
I think this needs to work inside builders - that library is designed to insulate the user from needing to know too much about the internals (i.e. it shouldn't really ever fail). I honestly don't think it needs too much work - it's one mod to that `_utils.py` file to make sure a suitable default is picked if the dtype is unsupported.

hgomersall:
(and a test of course  ;)

hgomersall:
@fredRos poke! I'd love to merge this if you can augment the builders... :smile:

fredRos:
Yes, I have a preliminary version that does the job but it needs a bit more polishing. I'd be glad if you had a look and told me if that's what you had in mind

test/test_pyfftw_numpy_interface.py
```diff
@@ -229,11 +229,12 @@ def _validate(self, array_type, test_shape, dtype,
 
         np_input_array = numpy.asarray(input_array)
 
-        if np_input_array.dtype == 'clongdouble':
-            np_input_array = numpy.complex128(input_array)
+        # TODO why are long double inputs copied to double precision?
```
hgomersall: I can't remember the precise reason for this. There are all sorts of annoyances around `clongdouble`, particularly on windows. I suspect this sidesteps one of those issues. It might be your work supersedes this approach.

test/test_pyfftw_partial.py
```diff
@@ -53,9 +55,13 @@ def test_failure(self):
             if dtype not in _supported_types:
                 a = empty_aligned((1,1024), npdtype, n=16)
                 b = empty_aligned(a.shape, dtype=a.dtype, n=16)
-                with self.assertRaisesRegex(NotImplementedError, "Rebuild pyFFTW with support for %s precision!" % _all_types_human_readable[dtype]):
+                msg = "Rebuild pyFFTW with support for %s precision!" % _all_types_human_readable[dtype]
+                with self.assertRaisesRegex(NotImplementedError, msg):
                     FFTW(a,b)
 
+    def test_conversion(self):
+        # If double precision not supported, the builder should convert to single precision
```
hgomersall: Are we assuming single precision is always supported? (also, you can put this as a docstring rather than a comment).

pyfftw/interfaces/numpy_fft.py
```diff
@@ -39,13 +39,15 @@
...
-array. Half precision input will be converted to single precision.  Otherwise,
-if any type conversion is required, the default will be double precision.
+It is notable that unlike :mod:`numpy.fftpack`, these functions will generally
+return an output array with the same precision as the input array, and the
+transform that is chosen is chosen based on the precision of the input array.
+That is, if the input array is 32-bit floating point, then the transform will
+be 32-bit floating point and so will the returned array. Half precision input
+will be converted to single precision. Otherwise, if any type conversion is
+required, the default will be double precision. If pyFFTW was not built with
+support for double precision, the default is long double precision. If that is not
```
hgomersall: Is that the right behaviour? I guess it is at least safe...

pyfftw/builders/_utils.py
```diff
@@ -60,18 +60,35 @@
...
-# precision.
+# Looking up a real dtype in here returns the complex complement of the same
+# precision, and vice versa.
 # It is necessary to use .char as the keys due to MSVC mapping long
 # double to double and the way that numpy handles this.
-_rc_dtype_pairs = {numpy.dtype('float32').char: numpy.dtype('complex64'),
+_rc_dtype_pairs = {}
+_default_dtype = None
+
+# Double precision is the default. Prefer casting to higher precision if
```
hgomersall: The default default :wink:

pyfftw/builders/_utils.py
```diff
@@ -119,22 +136,16 @@ def _Xfftn(a, s, axes, overwrite_input,
...
+        dtype_char = _default_dtype.char
+        if a.dtype == numpy.dtype('float16') and '32' in pyfftw._supported_types:
+            # convert half-precision to single precision, if available
+            dtype_char = numpy.dtype('float32').char
+        if not real or inverse:
+            # It's going to be complex
+            dtype_char = _rc_dtype_pairs[dtype_char].char
+
+        # convert the input array
+        a = numpy.asarray(a, dtype=dtype_char)
```
hgomersall: This is neater!

test/test_pyfftw_numpy_interface.py
hgomersall: I mean, if you're checking what types are available, that's probable enough!

test/test_pyfftw_partial.py
```diff
@@ -0,0 +1,72 @@
...
+    def test_failure(self):
+        for dtype, npdtype in zip(['32', '64', 'ld'], [np.complex64, np.complex128, np.clongdouble]):
+            if dtype not in _supported_types:
+                a = empty_aligned((1,1024), npdtype, n=16)
+                b = empty_aligned(a.shape, dtype=a.dtype, n=16)
+                msg = "Rebuild pyFFTW with support for %s precision!" % _all_types_human_readable[dtype]
+                with self.assertRaisesRegex(NotImplementedError, msg):
+                    FFTW(a,b)
+
+    def test_conversion(self):
```
hgomersall: Is the plan to expand this into an actual test?

hgomersall:
This looks to me to be pretty close. There is a question about necessary modifications to the dask interface test stuff. @jakirkham Are you able to comment on this? Possibly it's just necessary to fold the same changes in as in the numpy test case.

I've made a few inline comments.

test/test_pyfftw_numpy_interface.py
fredRos: Ok, I imagine this was supposed to work even if long double wasn't available? Weird, because `long double` is required but OK, I just remove the comments as we can now explicitly check the available types

test/test_pyfftw_partial.py
fredRos: I'm implementing the actual test, turned it into a docstring. And no, we don't assume single precision is always available, I changed the wording

pyfftw/interfaces/numpy_fft.py
fredRos: I don't lose precision unless it's inevitable. Perhaps we should emit a warning for narrowing conversions to alert the user?

pyfftw/builders/_utils.py
fredRos: Yes, I should make that my default default default

test/test_pyfftw_partial.py
fredRos: I have done so now

fredRos:
So there is one more problem in the tests with throwing the right exception. I want to fix it this week. But the automatic conversion in the builder now works and is tested. 

> There is a question about necessary modifications to the dask interface test stuff. @jakirkham Are you able to comment on this? Possibly it's just necessary to fold the same changes in as in the numpy test case.

I'm confused. Which changes are you talking about?

hgomersall:
[This Dask test stuff.](https://github.com/pyFFTW/pyFFTW/blob/master/test/test_pyfftw_dask_interface.py)

fredRos:
> So there is one more problem in the tests with throwing the right exception. I want to fix it this week. But the automatic conversion in the builder now works and is tested

After some thinking, I realize what the problem is. `scipy_fftpack` does not support `long double` and raises an exception. `numpy.fft` seems to support long double and doesn't raise an exception. In the tests, we use the same precisions as for numpy (see `test_pyfftw_scipy_interface.py`)

```python
if scipy.__version__ < '0.19':
    # Older scipy will raise an error for inputs of type float16, so we
    # cannot validate transforms with float16 input vs. scipy.fftpack
    complex_dtypes = pyfftw._supported_nptypes_complex
    real_dtypes = pyfftw._supported_nptypes_real
else:
    # reuse all dtypes from numpy tests (including float16)
    complex_dtypes = test_pyfftw_numpy_interface.complex_dtypes
    real_dtypes = test_pyfftw_numpy_interface.real_dtypes
```

The actual failure is that the original scipy and our plugin interface should raise the same exception but our interface doesn't raise one. I don't understand how this passed before.

I see two options:

1. we don't test long double on the scipy interface and accept that our interface behaves slightly differently. 
2. I add code that captures every call with `long double` in the interface. This will become a bit ugly because right now all calls end up in `_Xfftn` where the source interface of the call is not known currently.

I prefer option 1. Other precisions are not treated exactly the same, either. This is mentioned in the doc string in `numpy_fft.py`

```
It is notable that unlike :mod:`numpy.fftpack`, these functions will generally
return an output array with the same precision as the input array, and the
transform that is chosen is chosen based on the precision of the input array.
That is, if the input array is 32-bit floating point, then the transform will
be 32-bit floating point and so will the returned array. Half precision input
will be converted to single precision. Otherwise, if any type conversion is
required, the default will be double precision. If pyFFTW was not built with
support for double precision, the default is long double precision. If that is not
available, it defaults to single precision.
```

hgomersall:
I prefer option 1 too. Can't this still be included in the test with some suitable casting?

fredRos:
Here is how do it now in `test_pyfftw_scipy_interface.py`

```python
if scipy.__version__ < '0.19':
    # Older scipy will raise an error for inputs of type float16, so we
    # cannot validate transforms with float16 input vs. scipy.fftpack
    complex_dtypes = pyfftw._supported_nptypes_complex
    real_dtypes = pyfftw._supported_nptypes_real
else:
    # reuse all dtypes from numpy tests (including float16)
    complex_dtypes = test_pyfftw_numpy_interface.complex_dtypes
    real_dtypes = test_pyfftw_numpy_interface.real_dtypes

# Remove long double because scipy explicitly doesn't support it
complex_dtypes = [x for x in complex_dtypes if x != numpy.clongdouble]
real_dtypes = [x for x in real_dtypes if x != numpy.longdouble]
```

It seems to work on linux w/o any dask modifications. One build failed I guess for network problems. However, the wisdom export on windows fails. No idea why, I didn't touch this at all

hgomersall:
It was more that the interface tests should include the Dask case too. I think the tweaks the numpy interface test code need to be extended to the dask tests.

fredRos:
> I think the tweaks the numpy interface test code need to be extended to the dask tests.
Ok, I will look into this that.

I have no idea why only the wisdom tests fail, and only on windows. @hgomersall Please have a look at https://ci.appveyor.com/project/hgomersall/pyfftw/build/1.0.398/job/aepag9u5cn1pv84s for example. The wisdom seems to be unchanged when creating a plan. Perhaps it's something as obscure as in #3. I looked back and the failure appeared pretty much from the beginning of my branch. Diffing `export_wisdom` in `pyfftw.pyx` to the master branch, I only reordered the commands inside `IF HAVE_DOUBLE:` etc. The only difference I can see is that I initialize

```cython
const char* empty = ''
bytes py_wisdom  = empty
...
```

fredRos:
> I think the tweaks the numpy interface test code need to be extended to the dask tests.

I checked, the tests work automatically because the data types are taken over from the numpy test. In `test_pyfftw_dask_interface.py`

```python
from .test_pyfftw_numpy_interface import complex_dtypes, real_dtypes
...
```
and all the conversion logic is taken over automatically, too. So no work required. I will take over @grlee77 latest fixes by rebasing to master

grlee77:
In regards to the scipy interface tests not failing previously with long double inputs (I think you may have already figured this out):

`scipy.fftpack.*` will raise errors on long-double input, but the lines below explicitly cast these inputs down to double precision prior to calling scipy routines so that no error will be raised.  (This conversion is done for `numpy.fft.*` as well, although at least on recent numpy this does not seem to be necessary.  I can call np.fft.fft with float128 input and it runs just fine, silently converting to double precision.)
https://github.com/pyFFTW/pyFFTW/blob/ad23f73defdb2ed27c834cadeff2fd772953a8a6/test/test_pyfftw_numpy_interface.py#L223-L227

The long double precision inputs are still used in the FFTW interfaces.  If we just remove the long double test cases as you have done in 135ece5 , then we would not be testing our scipy FFTW interfaces for those dtypes.  I think we want to continue allowing the testing of these precisions in the interfaces even though the comparison to the underlying scipy.fftpack routines is actually based on double-precision.

grlee77:
sorry, after reading your comments again, I see that you primarily want to remove the long double test cases so that the expected exception types match.  I am okay with that.

I guess a third option would be to make the SciPy interfaces raise a `ValueError` on long double input to match the `scipy.fftpack` behaviour, but that is probably worse as it is backwards-incompatible with prior pyFFTW releases and we are already not trying to exactly match dtypes in the numpy interfaces either  (i.e. `numpy.fft` always uses double precision regardless of input dtype)

grlee77:
This is looking pretty good to me.  I think the final test failure is because for Python 3.3, a version of numpy older than 1.10 is probably being used.  The [cast to double precision within numpy.fft functions](https://github.com/numpy/numpy/blob/v1.14.0/numpy/fft/fftpack.py#L189) was not present in 1.9.3 and earlier. 

I would be in favor of just bumping the minimum numpy required to 1.10 as that is now more than two years old (Oct, 5 2015 release date).  Otherwise, we will need to restore the lines in the numpy `_validate` methods that explicitly cast the input to double precision.

test/test_pyfftw_numpy_interface.py
```diff
@@ -229,14 +229,6 @@ def _validate(self, array_type, test_shape, dtype,
 
         np_input_array = numpy.asarray(input_array)
 
-        # TODO why are long double inputs copied to double precision?
```
grlee77: This appears to have been necessary for numpy versions older than 1.10.  Starting from 1.10, each numpy.fft function always copies the input to double precision so that it no longer has to be done here.

hgomersall:
@fredRos To be clear, the problem with the wisdom test failures is that `forget_wisdom` is not working. Does that cover it?

hgomersall:
@grlee77 Happy to drop support for older Scipy. Does this mean we drop support for 3.3? Also not too problematic I'd imagine?

grlee77:
> Happy to drop support for older Scipy.

I think you meant Numpy (bump requirement to 1.10)?  We could, bump SciPy as well, but I don't think there is a strong reason to go to a particular version.  Some float16 handling was updated in 0.19, but that was pretty recent so probably want to still keep the support for older versions around as well.

For Python, I agree that we should bump to 3.4 (that is the minimum version that numpy and scipy currently test against and build wheels for).  pyFFTW should still work on older versions, we just wouldn't guarantee that the tests will all pass. 

We can make a separate PR to bump the various minimum versions, but I think either the numpy bump or restoring the code casting the inputs to np.fft.* to double precision will have to occur in this PR to get all CI to pass.

hgomersall:
@grlee77 Yes, that's reasonable - weak support is fine for numpy and python <3.4.

Does the version issue fix the wisdom import/export test failures?

grlee77:
I didn't see any wisdom related failures in the CI results for the most recent commit.  There was only a failure on 3.3 at a lot of places like this: https://travis-ci.org/pyFFTW/pyFFTW/jobs/326863951#L904 which I suspect are probably due to an older numpy getting used on 3.3.

hgomersall:
@grlee77 it's in [Appveyor](https://ci.appveyor.com/project/hgomersall/pyfftw).

grlee77:
> @grlee77 it's in Appveyor.

Right, I had not looked at the Wisdom comments above in detail.  No idea why that is occurring, but it is for all Python versions, so dropping 3.3 support isn't going to help there.

fredRos:
> We can make a separate PR to bump the various minimum versions, but I think either the numpy bump or restoring the code casting the inputs to np.fft.* to double precision will have to occur in this PR to get all CI to pass.

I restored the casting to 64 bits. So the tests run as before, even for scipy. You can bump  versions in a separate PR. 

> I have no idea why only the wisdom tests fail, and only on windows

I still don't know why the wisdom is not updated on appveyor in contrast to travis but it may have something to do with the hardware that FFTW detects on those build systems. I realized that in `master`, the wisdom updating is **not** tested for long double (why?) but I do in my tests; see  https://github.com/pyFFTW/pyFFTW/blob/master/test/test_pyfftw_wisdom.py#L83

```python 
for n in range(0,2):
    self.assertNotEqual(before_wisdom[n], after_wisdom[n])
```

I now skip the test for long double on windows https://github.com/pyFFTW/pyFFTW/pull/189/commits/3ca42dfcbeaf1d7ed29b99083869a04bc04d65bc#diff-895bdd25e71eb2d8b71c52b2ca54dae5R63

```python
            if prec == 'ld' and sys.platform.startswith("win"):
                pass
            else:
                self.assertNotEqual(before, after)
```
The tests on appveyor are queued but I'm optimistic they will pass. If they do, it's ready to merge from my side

hgomersall:
@fredRos I don't think we should stall merging this for the corner cases you've identified, but the skips should be explicit: `raise SkipTest('reason')`.

I think I saw a couple of other places where you've skipped a test quietly that can be done with an explicit `raise`. IMO, if we're skipping tests we should be upfront about why so everyone knows and we can fix it later.

fredRos:
Yes, I agree to be upfront. But in this case I don't want to skip the whole test because it's just one of several asserts, have a look at https://github.com/pyFFTW/pyFFTW/pull/189/files#diff-895bdd25e71eb2d8b71c52b2ca54dae5

Anyways, do you remember why long double was not at all tested before on the wisdom export?

fredRos:
Just to add, all appveyor tests passed, too! :smile:

hgomersall:
:smile: Great!

Oh I see. One way to solve this is to have a different test class for each dtype. This can be build dynamically like in the [scipy tests](https://github.com/pyFFTW/pyFFTW/blob/ad23f73defdb2ed27c834cadeff2fd772953a8a6/test/test_pyfftw_scipy_interface.py#L186).

hgomersall:
I think long double wasn't tested explicitly because I assumed the wisdom framework would be dtype agnostic. Apparently I was wrong!

fredRos:
@hgomersall ping, is there anything else I have to do before you merge? I saw that my changes already helped on rasberry pi in #203

hgomersall:
@fredRos Nothing I think. Have we now removed (swept under the carpet) the problem with Appveyor?

hgomersall:
Ah yes, you said that.

hgomersall:
I think we should add an issue about the appveyor tests, then we can merge.

fredRos:
I'm not sure it has something to do with windows, it may just be the hardware. Right now I don't test if running a long double FFT updates the wisdom. It feels like it should but it doesn't. However, there is no mention in the FFTW docs that the wisdom **must change**. It simply may change, that's why I feel fine about skipping the test

hgomersall:
I agree. I'm happy to merge this. I had a comment about protection of the NULL functions, but basically it looks fine.

hgomersall:
@fredRos how are the null function handled?

fredRos:
Can you please clarify, what null function do you mean?

hgomersall:
My interpretation of your changes is that now when a type is not available, the function pointer is set to null in the lookup tables. I just couldn't see a defensive catch for that anywhere, but it might be unnecessary.

fredRos:
An informative exception is raised in case an unsupported precision is requested!

All access to possibly NULL function pointers is through `scheme_functions` in
```cython
def scheme_functions(scheme):
    if scheme in _scheme_functions:
        return _scheme_functions[scheme]
    else:
        msg = "The scheme '%s' is not supported." % str(scheme)
        if scheme[1] in _all_types:
            msg += "\nRebuild pyFFTW with support for %s precision!" % \
                   _all_types_human_readable[scheme[1]]
        raise NotImplementedError(msg)
```

and used inside `cdef class FFTW` as:

```cython
        ...
        functions = scheme_functions(scheme)

        self._fftw_planner = planners[functions['planner']]
        self._fftw_execute = executors[functions['executor']]
        self._fftw_destroy = destroyers[functions['generic_precision']]
        ...
```

hgomersall:
That looks fine. Can we add a quick defensive catch on each of those functions? Like on [Line1140](https://github.com/pyFFTW/pyFFTW/blob/master/pyfftw/pyfftw.pyx#L1140), something like RuntimeError("A planner function was expected but returned as a NULL function. This is a bug.").

This is purely defensive against future potential changes that don't necessarily factor in everything that has gone before.

hgomersall:
To be clear, everything else looks good from my end.

fredRos:
Check if you like this

hgomersall:
Looks great!

hgomersall:
@grlee77 It would be good to have a sanity check on this before we merge if you have time, but I think it looks fine.

grlee77:
Sorry, I didn't get back to this earlier.  It looks good to me as well.  There may be a few commits here and there that could be squashed, but mostly they are pretty clean and I am fine with merging it as is.  Great work on this!

Once merged, I will try things out with `conda-build` again to make sure everything is still working there as well.  Then, once we make a release we should be able to add Windows pyFFTW packages on `conda-forge` as well.

jakirkham:
Sorry haven't had time to look into this. The Dask interface stuff is tested pretty well in Dask against `numpy.fft`. So as long as our `numpy_fft` interface remains largely the same, Dask won't care too much. At some point I would like to vastly simplify how we are testing the Dask interface here as I don't think we need to be testing it to the same degree as we are testing `numpy_fft`.

fredRos:
Thank  you very much. This has taken a looong time but the code quality definitely increased through the reviews. Next time we should be a bit faster. 

I can now work on merging the last piece, the actual reason why I added this and the previous #177 : FFTW with MPI. Stay tuned...

grlee77:
Building worked like a charm with the FFTW libs from `conda-forge`.  I did encounter two test failures that I have attempted to fix in #209.  These appear to be because numpy's `longdouble` is `double` on the system where I did the testing.

PR #190, opened 2017-09-03
WIP: support building from conda-forge FFTW packages on Windows

grlee77:
This PR adds two environment variables that allow the assumed library path and filenames to be modified to work with conda-forge's FFTW packages on Windows.  The existing Windows installation behavior is unmodified when these environment variables are not defined.

1.  Unlike the pre-built binaries from fftw.org, the conda-forge libraries have names like `fftw3f.dll` rather than `libfftw3f-3.dll`.  This naming convention can be selected by defining `PYFFTW_WIN_CONDAFORGE`.  

2. `PYFFTW_LIB_DIR` should be set to the folder containing `fftw3.lib` and `fftw3f.lib`.  For conda-forge this is the `lib` subfolder of the environment.

3.  On conda, the dll files will already be on the path (via the `\bin` subfolder of the environment), so they do not need to be included in `package_data`.  

**Note:** conda-forge's Windows build does not include long-double support, so the build will only complete once this PR is combined with the proposed changes from #189.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -96,8 +96,14 @@ def get_package_data():
     package_data = {}
 
     if get_build_platform() in ('win32', 'win-amd64'):
-        package_data['pyfftw'] = [
-            'libfftw3-3.dll', 'libfftw3l-3.dll', 'libfftw3f-3.dll']
+        if 'PYFFTW_WIN_CONDAFORGE' in os.environ:
+            # fftw3.dll, fftw3f.dll will already be on the path (via the
+            # conda environment's \bin subfolder)
+            pass
+        else:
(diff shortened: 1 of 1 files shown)


hgomersall:
Apologies for the delay on this. It seems pretty trivial and easy to merge. Does it interact ok with #198?

grlee77:
Thanks!

The conda-forge packages are built on the latest official release, so this won't get tested immediately.  I will make a [DO NOT MERGE] PR over at the conda-forge feedstock to verify that it passes CI when building from current master.  That way any potential unanticipated issues can be worked out prior to release.

Issue #136, opened 2016-10-01
RuntimeWarning: overflow encountered in long_scalars


david-hoffman:
I've been getting the following error when attempting to perform an fft on a large array:

``` console
C:\Anaconda3\lib\site-packages\pyfftw\builders\_utils.py:127: RuntimeWarning: overflow encountered in long_scalars
  output_array = pyfftw.empty_aligned(output_shape, output_dtype)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-65-12bb74fa6d42> in <module>()
----> 1 pyfftw.fftn(a);

C:\Anaconda3\lib\site-packages\pyfftw\interfaces\numpy_fft.py in fftn(a, s, axes, norm, overwrite_input, planner_effort, threads, auto_align_input, auto_contiguous)
    181     return _Xfftn(a, s, axes, overwrite_input, planner_effort,
    182             threads, auto_align_input, auto_contiguous,
--> 183             calling_func, normalise_idft=normalise_idft, ortho=ortho)
    184
    185 def ifftn(a, s=None, axes=None, norm=None, overwrite_input=False,

C:\Anaconda3\lib\site-packages\pyfftw\interfaces\_utils.py in _Xfftn(a, s, axes, overwrite_input, planner_effort, threads, auto_align_input, auto_contiguous, calling_func, normalise_idft, ortho)
    123         planner_args = (a, s, axes) + args
    124
--> 125         FFTW_object = getattr(builders, calling_func)(*planner_args)
    126
    127         # Only copy if the input array is what was actually used

C:\Anaconda3\lib\site-packages\pyfftw\builders\builders.py in fftn(a, s, axes, overwrite_input, planner_effort, threads, auto_align_input, auto_contiguous, avoid_copy)
    362     return _Xfftn(a, s, axes, overwrite_input, planner_effort,
    363             threads, auto_align_input, auto_contiguous,
--> 364             avoid_copy, inverse, real)
    365
    366 def ifftn(a, s=None, axes=None, overwrite_input=False,

C:\Anaconda3\lib\site-packages\pyfftw\builders\_utils.py in _Xfftn(a, s, axes, overwrite_input, planner_effort, threads, auto_align_input, auto_contiguous, avoid_copy, inverse, real)
    125         a_copy = a.copy()
    126
--> 127     output_array = pyfftw.empty_aligned(output_shape, output_dtype)
    128
    129     flags = [planner_effort]

[...]

PR #192, opened 2017-09-16
simple fix to overflow in empty_aligned() (on Windows machine)

hyliu1989:
The previous codes (even with the explicit multiplication in lines 196 to 198 in utils.pxi) still depend on the machine default which causes pyFFTW on Windows machine to throw overflow error. It fixes the problem to specify the type of variable `array_length`.

### DIFF
--- a/pyfftw/utils.pxi
+++ b/pyfftw/utils.pxi
@@ -182,6 +182,8 @@ cpdef empty_aligned(shape, dtype='float64', order='C', n=None):
     determine alignment. The rest of the arguments are as per
     :func:`numpy.empty`.
     '''
+    cdef long long array_length
+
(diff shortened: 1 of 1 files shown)


hgomersall:
I'm sorry I missed this. The CI tests seemed to failed so I've triggered another attempt - it's not obvious to me why it should have done!

grlee77:
this is a fix for #136, but ideally could still use a test case, so I have left the issue open for now.

hgomersall:
Yeah I agree in principle. I merged without the test case on the basis it didn't actually change behaviour. It's not totally trivial to submit a test case because it by nature only occurs on very large arrays on certain systems.

Issue #197, opened 2017-11-19
Numpy interface corrupts array argument


bwohlberg:
There appears to be a platform-dependent bug in the numpy interface that results in corruption of the array passed to ``pyfftw.interfaces.numpy_fft.rfftn``. Under 32 bit Ubuntu 16.04 running on an Intel T2500 CPU, the following script

```
from __future__ import print_function

import hashlib
import numpy as np
import pyfftw

pyfftw.interfaces.cache.enable()

def rfftn(a, s=None, axes=None):
    return pyfftw.interfaces.numpy_fft.rfftn(a, s=s, axes=axes,
                overwrite_input=False, planner_effort='FFTW_MEASURE')

def hash(x):
    return hashlib.md5(x).hexdigest()[0:8]

np.random.seed(12345)
N = 64
U = np.random.randn(N, N).astype(np.float32)
V = np.random.randn(N, N).astype(np.float32)

for k in range(3):
    print(hash(U), hash(V))
    x = rfftn(U, None, axes=(0,1))
    y = rfftn(V, None, axes=(0,1))
```

gives output

```
a7662abd bb54ccca
a7662abd bb54ccca
a7662abd a7662abd
```

or

```
a7662abd bb54ccca
bb54ccca bb54ccca
bb54ccca bb54ccca
```

indicating that calling ``pyfftw.interfaces.numpy_fft.rfftn`` with ``U`` and then ``V`` results in ``U`` overwriting ``V`` or vice-versa.

This only seems to occur when the script is run with Python 3.5 (``python3`` on Ubuntu 16.04), and does not occur when the initial call to ``pyfftw.interfaces.cache.enable()`` is removed.

hgomersall:
I can't replicate this on my machine (16.04 Ubuntu but 64-bit) (I just get the . Are you able to pin down the bug better? It is probably happening in [pyfftw.interfaces._utils](https://github.com/pyFFTW/pyFFTW/blob/master/pyfftw/interfaces/_utils.py), with arrays not being protected properly.

Thanks!

bwohlberg:
From the attached debugging code (a modified version of the original
bug demonstration script, a version of pyfftw.interfaces._utils with
added debugging code, and the output of the script), it appears that
the corruption is occurring in the final "FFTW_object(input_array=a,
...)" call in _utils.py, and affects the "other" array, i.e., the
array to which rfftn was previously applied, not the argument a
itself.

Does that help?


[...]

PR #198, opened 2017-12-21
Fix broken build with OpenMP

grlee77:
Apparently HAVE_DOUBLE_THREADS, etc. were never set to either True or False when the OpenMP libraries were found instead.  This PR should fix that.  This was a source of recent build failures as in #192 (see: https://travis-ci.org/pyFFTW/pyFFTW/jobs/276350807#L899) and as mentioned by a user in #197.

I am not sure how this was previously passing on Travis CI, but perhaps the set of FFTW libs that gets installed on Travis has changed in the time since #189 was merged.

@fredRos, do you mind taking a quick look at the proposed change?

### DIFF
--- a/pyfftw/interfaces/__init__.py
+++ b/pyfftw/interfaces/__init__.py
@@ -239,10 +239,12 @@ else:
 try:
     import dask.array.fft
     dask.array.fft.fft_wrap
-except ImportError:
+except (ImportError, AttributeError):
     pass
-except AttributeError:
-    del dask
 else:
-    del dask
     from . import dask_fft
+
+try:
+    del dask
+except NameError:
+    pass
--- a/setup.py
+++ b/setup.py
@@ -257,6 +257,9 @@ class EnvironmentSniffer(object):
                 lib_omp = False
                 self.compile_time_env[self.HAVE(d, 'OMP')] = False
 
+            if lib_omp:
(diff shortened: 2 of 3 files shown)


grlee77:
This appears to have resolved the TravisCI failures. 

There appear to be other unrelated failures on Appveyor.  One appears to be related to not having dask installed:
https://ci.appveyor.com/project/hgomersall/pyfftw/build/1.0.381/job/sv5bggx337nete5q#L633

grlee77:
actually, in that Appveyor failure, the `dask` module itself is raising an AttributeError on numpy.cbrt:
https://ci.appveyor.com/project/hgomersall/pyfftw/build/1.0.381/job/is15q0njjpp0pmrk#L629

Perhaps something is broken in the dask `conda` package for some configurations.

grlee77:
I now understand why the `del dask` raises an error.  The logic in the code assumes the `AttributeError` is raised by dask.array.fft no having the desired method.  In this case, the attribute error is raised by numpy before dask import is complete so it is not possible to delete the `dask` module.

I guess we would need something like the following for things to work properly in that case:
```Python
try:
    del dask
except NameError:
    pass
```

grlee77:
Appveyor is now passing as well

hgomersall:
@grlee77 Thanks for the very necessary fix here!

PR #199, opened 2017-12-22
Fix to #197 in which misaligned arrays in the cache caused overwrites…

hgomersall:
… of apparently unrelated arrays.

The fix is to have the alignment of the array as part of the key for the cache lookup.

We use the `pyfftw.simd_alignment` attribute as we can assume that's the most stringent alignment case.

### DIFF
--- a/test/test_pyfftw_interfaces_cache.py
+++ b/test/test_pyfftw_interfaces_cache.py
@@ -336,7 +336,8 @@ class CacheTest(unittest.TestCase):
         Um = pyfftw.empty_aligned((N, N+1), dtype=np.float32, order='C')
         Vm = pyfftw.empty_aligned((N, N+1), dtype=np.float32, order='C')
         U = np.ndarray((N, N), dtype=Um.dtype, buffer=Um.data, offset=0)
-        V = np.ndarray((N, N), dtype=Vm.dtype, buffer=Vm.data, offset=4)
+        V = np.ndarray(
+            (N, N), dtype=Vm.dtype, buffer=Vm.data, offset=Vm.itemsize)
 
         U[:] = np.random.randn(N, N).astype(np.float32)
         V[:] = np.random.randn(N, N).astype(np.float32)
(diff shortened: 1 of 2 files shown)


hgomersall:
@grlee77 @bwohlberg are you able to give this a PR quick sanity check? It's with reference to #197.

test/test_pyfftw_interfaces_cache.py
```diff
@@ -318,6 +321,39 @@ def test_objects_removed_after_keepalive(self):
...
+        interfaces.cache.enable()
+
+        N = 64
+        pyfftw.interfaces.cache.enable()
+        np.random.seed(12345)
+
+        Um = pyfftw.empty_aligned((N, N+1), dtype=np.float32, order='C')
+        Vm = pyfftw.empty_aligned((N, N+1), dtype=np.float32, order='C')
+        U = np.ndarray((N, N), dtype=Um.dtype, buffer=Um.data, offset=0)
+        V = np.ndarray((N, N), dtype=Vm.dtype, buffer=Vm.data, offset=4)
```
grlee77: maybe change from `offset=4` to `offset=Vm.itemsize`?

test/test_pyfftw_interfaces_cache.py
```diff
@@ -318,6 +321,39 @@ def test_objects_removed_after_keepalive(self):
...
+
+        U[:] = np.random.randn(N, N).astype(np.float32)
+        V[:] = np.random.randn(N, N).astype(np.float32)
+
+        uh = hashlib.md5(U).hexdigest()
+        vh = hashlib.md5(V).hexdigest()
+        x = interfaces.numpy_fft.rfftn(
+            U, None, axes=(0,1), overwrite_input=False)
```
grlee77: PEP8 -> `axes=(0, 1)` here and a couple lines below

hgomersall:
Thanks, both helpful!

bwohlberg:
This solves the problem for me, including on the original 32 bit architecture where it first became apparent.

PR #200, opened 2017-12-28
update version import syntax in doc/conf.py for compatibility with versioneer

grlee77:
I missed updating the 'conf.py` version import syntax in my versioneer PR (#176).  This change should fix the `Readthedocs` builds and get the badge on the home page green again.


grlee77:
I think readthedocs only runs on master, so I don't think we will be able to verify the fix until after this is merged.  It does work when I tried it locally.

jakirkham:
Sorry, should have pointed this out in the versioneer PR. In any event, glad to see this change land.

Issue #128, opened 2016-06-07
Testsuite failures on selected architectures due to alignment problem


ghisvail:
I have updated the [Debian packaging](https://buildd.debian.org/status/package.php?p=pyfftw) with release 0.10.4 and the build now fails for some architectures, some of these considered major. On arm64, powerpc, ppc64el and s390x, the failing tests seem all related to alignment. If the package is intended to be not portable, please acknowledge it here and I'll restrict the build to the qualifying architectures.

Thanks,
Ghis

hgomersall:
Ach, I've no idea why it's suddenly broken. The package is intended to be portable, but I've no mechanism for testing against anything but x86/x86-64. I know the alignment code for ARM is not optimised. What is the failure?

ghisvail:
> What is the failure?

You can click on individual `Build-Attempted` fields [here](https://buildd.debian.org/status/package.php?p=pyfftw) to read the full build log. All seems to contain this:
```
======================================================================
ERROR: test_get_alignment (test.test_pyfftw_utils.UtilsTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "test/test_pyfftw_utils.py", line 84, in test_get_alignment
    if 'avx' in each_cpu['flags']:
KeyError: 'flags'

======================================================================
FAIL: test_alignment (test.test_pyfftw_complex.Complex128FFTWTest)
Test to see if the alignment is returned correctly
----------------------------------------------------------------------
Traceback (most recent call last):
  File "test/test_pyfftw_complex.py", line 293, in test_alignment
    self.assertTrue(fft.input_alignment == 16)
AssertionError: False is not true

======================================================================
FAIL: test_alignment (test.test_pyfftw_complex.Complex64FFTWTest)
Test to see if the alignment is returned correctly
----------------------------------------------------------------------
Traceback (most recent call last):
  File "test/test_pyfftw_complex.py", line 293, in test_alignment
    self.assertTrue(fft.input_alignment == 16)
AssertionError: False is not true

======================================================================
FAIL: test_alignment (test.test_pyfftw_real_backward.RealBackwardDoubleFFTWTest)
Test to see if the alignment is returned correctly
----------------------------------------------------------------------
Traceback (most recent call last):
  File "test/test_pyfftw_complex.py", line 293, in test_alignment
    self.assertTrue(fft.input_alignment == 16)
AssertionError: False is not true

======================================================================
[...]

PR #201, opened 2017-12-28
Fix test suite on machines without any SIMD instruction sets

grlee77:
This is an attempt to address the ARM test failures reported by @ghisvail in #128.  We currently only support SSE and AVX SIMD variants.  If neither is found, a few tests related to SIMD auto-alignment are skipped.

### DIFF
--- a/test/test_pyfftw_builders.py
+++ b/test/test_pyfftw_builders.py
@@ -36,6 +36,7 @@ from pyfftw import builders, empty_aligned, byte_align, FFTW
 from pyfftw.builders import _utils as utils
 from .test_pyfftw_base import run_test_suites
 from ._get_default_args import get_default_args
+from pyfftw.pyfftw import _valid_simd_alignments
 
 import unittest
 import numpy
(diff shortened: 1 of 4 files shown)


hgomersall:
Can we be more sensible about the way we infer the ARM alignment? It seems to me that all that is really needed is to fix the alignment stuff on ARM and then make sure the build system works properly.

The issue so far is the lack of ARM CI, though perhaps this exists and I'm just ignorant of it.

grlee77:
From what I can tell, the current code is querying whether the instruction sets are supported by the CPU and setting alignment based on that.  However, the FFTW libraries used may not have been built with a particular SIMD instruction set enabled.  It seems that our current code would, for example, always enforce alignment for AVX on newer intel x86_64 processors, but if the underlying FFTW wasn't built with `--enable-avx` then this is not actually providing any benefit?  

I am wondering if there is some way to query FFTW itself for what alignment is optimal?  I see that there is a function called [fftw_malloc](http://www.fftw.org/fftw3_doc/Memory-Allocation.html) that can be used to allocate memory with an optimal SIMD alignment, but I'm not sure if there is some simple way to query FFTW for what that alignment is.  If I get a chance I can try to see how it is being determined in the FFTW source.

The conda-forge packages are currently built with support for SSE, SSE2, and AVX only.

For ARM, support for NEON SIMD instructions was added in FFTW 3.3.1.  It seems it was initially 64-bit alignment, but was updated to use 128-bit alignment in 3.3.3 ([release notes](http://www.fftw.org/release-notes.html)) .  This requires building FFTW with `--enable-neon`.

hgomersall:
There is little harm in enforcing a more stringent alignment than FFTW supports (as in, it's safe). I'm ok with the potential performance hit that might occur with some machines that don't support AVX in the lib but do on the CPU. I feel the ease of referencing the core outweighs the potential corner case issues.

We should support ARM properly, but I don't know how to query the CPU for the alignment (I've not really looked). I've requested some resources on [WorksOnArm](https://github.com/WorksOnArm) so perhaps we can get the CI up and running on ARM.

grlee77:
We could detect if we are on ARM at compile time using something like:
```C
#if defined(_MSC_VER)
#ifdef _M_ARM
#define ARCH_ARM
#endif
#endif

#if defined(__GNUC__)
#ifdef __arm__
#define ARCH_ARM
#endif
#endif
```
and if ARM is detected, enforce 128-bit alignment.  This would not be detecting if FFTW was compiled with NEON support, but should at least be providing a compatible alignment if it was.

hgomersall:
That works for me. Are you able to add it to this patchset?

grlee77:
Yeah, I can do it later today.

grlee77:
I implemented the changes in that last commit here, but have not tested it on any actual ARM hardware.  Did you hear anything back from WorksOnArm?

hgomersall:
Yeah, they're not ready for running a travis like environment :disappointed: . It seems possible to test under qemu, but that seems nasty and I suspect we don't have all out lives to devote to this. I could set up RPi on a static IP as a test box...?

hgomersall:
Should this be good to go now as well?

hgomersall:
(aside from the lack of testing box)

grlee77:
Yeah, I think it is good to go.  If we find time to test on ARM later we can always revisit if there are still any other unanticipated issues.

PR #204, opened 2018-01-16
Update minimum requirements

grlee77:
All minimum versions selected here are at least 2 years old.

Our current `setup.py` seemed to already be enforcing Cython >=0.23, so that is what I went with elsewhere as well.

The choice of scipy 0.16 is somewhat arbitrary.  I just selected a version that is a bit more than 2 years old at this point.

Numpy 1.10 corresponds to a version where automatic casting of inputs to the `numpy.fft.*` functions to double precision was implemented.  We could potentially drop our own manual casting of the inputs to double within the numpy interface tests after this PR is merged.

Python 3.3 was being tested on Travis only.  I have removed it and added 3.6 there instead.  Appveyor was already using 3.4, 3.5 and 3.6, but I have tried to make sure that some test cases get run with the minimum required versions and then have 3.6 using more recent versions.  

~The new `pyproject.toml` is for [PEP 518](https://www.python.org/dev/peps/pep-0518/) support which has been [implemented in pip](https://github.com/pypa/pip/pull/4144).  I based this off of the one currently in scipy, with the minimum numpy bumped up to 1.10 to match our new minimum requirement.  Basically you want as old of a numpy as possible for a given python version to take advantage of ABI backwards compatibility (see much discussion: [here](https://github.com/pypa/pip/issues/4582#issuecomment-315037686)).~
**edit**: pip is not quite ready yet for pyproject.toml.  Can be added at some point in the future via #226

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,9 +9,9 @@ env:
 
 python:
   - 2.7
-  - 3.3
   - 3.4
   - 3.5
+  - 3.6
 
 branches:
     except:
--- a/README.rst
+++ b/README.rst
@@ -48,13 +48,12 @@ index page is `here <http://pypi.python.org/pypi/pyFFTW>`_.
 
 Requirements (i.e. what it was designed for)
 --------------------------------------------
-- Python 2.7 or greater (Python 3 is supported)
-- Numpy 1.6
+- Python 2.7 or 3.4+
+- Numpy 1.10
 - FFTW 3.3 or higher (lower versions *may* work) libraries for single, double,
   and long double precision in serial and multithreading (pthreads or openMP)
-  versions
-- Cython 0.15 or higher (though the source release on PyPI loses this
-  dependency)
+  versions.
+- Cython 0.23 or higher
 
 (install these as much as possible with your preferred package manager).
 
--- a/README.rst
+++ b/README.rst
@@ -23,8 +23,9 @@
 PyFFTW
 ======
 
-pyFFTW is a pythonic wrapper around `FFTW 3 <http://www.fftw.org/>`_, the
-speedy FFT library.  The ultimate aim is to present a unified interface for all the possible transforms that FFTW can perform.
+pyFFTW is a pythonic wrapper around FFTW_ 3, the speedy FFT library.  The
+ultimate aim is to present a unified interface for all the possible transforms
+that FFTW can perform.
 
 Both the complex DFT and the real DFT are supported, as well as on arbitrary
 axes of abitrary shaped and strided arrays, which makes it almost
@@ -38,24 +39,33 @@ Operating FFTW in multithreaded mode is supported.
 pyFFTW implements the numpy and scipy fft interfaces in order for users to
 take advantage of the speed of FFTW with minimal code modifications.
 
-A comprehensive unittest suite can be found with the source on the github
-repository or with the source distribution on PyPI.
+A comprehensive unittest suite can be found with the source on the GitHub_
+repository or with the source distribution on PyPI_.
 
-The documentation can be found on
-`Read the Docs <http://pyfftw.readthedocs.io>`_, the source is
-on `github <https://github.com/pyFFTW/pyFFTW>`_ and the python package
-index page is `here <http://pypi.python.org/pypi/pyFFTW>`_.
+The documentation can be found on `Read the Docs`_, the source is on GitHub_
+and the python package index page (PyPI_).  Issues and questions can be
+raised at the `GitHub Issues`_ page.
 
 Requirements (i.e. what it was designed for)
 --------------------------------------------
-- Python 2.7 or 3.4+
-- Numpy 1.10
-- FFTW 3.3 or higher (lower versions *may* work) libraries for single, double,
+- Python_ 2.7 or >= 3.4
+- Numpy_ >= 1.10.4  (lower versions *may* work)
+- FFTW_ >= 3.3 (lower versions *may* work) libraries for single, double,
   and long double precision in serial and multithreading (pthreads or openMP)
   versions.
-- Cython 0.23 or higher
+- Cython_ >= 0.23 (lower versions *may* work)
 
 (install these as much as possible with your preferred package manager).
+
+In practice, pyFFTW *may* work with older versions of these dependencies, but
+it is not tested against them.
+
+Optional Dependencies
+---------------------
+- Scipy_ >= 0.16
+- Dask_ >= 0.14.2
+
+Scipy and Dask are only required in order to use their respective interfaces.
 
 Installation
 ------------
@@ -69,15 +79,14 @@ or::
 
   easy_install pyfftw
 
-Installers are on the PyPI page for both 32- and 64-bit Windows, which include
+Installers are on the PyPI_ page for both 32- and 64-bit Windows, which include
 all the necessary DLLs.
 
 With FFTW installed, the PyPI release should install fine on Linux and Mac OSX. It doesn't mean it won't work anywhere else, just we don't have any information on it.
 
-Windows development builds are also automatically uploaded to
-`bintray <https://bintray.com/hgomersall/generic/PyFFTW-development-builds/view>`_
-as wheels (which are built against numpy 1.9), from where they can be
-downloaded and installed with something like::
+Windows development builds are also automatically uploaded to bintray_ as
+wheels (which are built against numpy 1.9), from where they can be downloaded
+and installed with something like::
 
   pip install pyFFTW-0.10.0.dev0+79ec589-cp35-none-win_amd64.whl
 
@@ -112,7 +121,7 @@ option is available, pyFFTW works in serial mode only.
 
 For more ways of building and installing, see the
 `distutils documentation <http://docs.python.org/distutils/builtdist.html>`_
-and `setuptools documentation <https://pythonhosted.org/setuptools/>`_.
+and `setuptools documentation <https://setuptools.readthedocs.io>`_.
 
 Platform specific build info
 ----------------------------
@@ -155,13 +164,8 @@ Now install pyfftw from pip::
 
   pip install pyfftw
 
-Notes: `pkgin <http://saveosx.org>`_ fftw package does not contain the long or
-float implementations of fftw and so pyFFTW can only perform double-precision
-transforms.
-
-It has been suggested that `macports <http://www.macports.org/>`_ might also
-work fine. You should then replace the LD environmental variables above with the
-right ones.
+It has been suggested that macports_ might also work fine. You should then
+replace the LD environmental variables above with the right ones.
 
 - DYLD - path for libfftw3.dylib etc - ``find /usr -name libfftw3.dylib``
 - LDFLAGS - path for fftw3.h - ``find /usr -name fftw3.h``
@@ -191,7 +195,7 @@ See some of my philosophy on testing in development `here
 If you want to argue with the philosophy, there is probably a good place to
 do it.
 
-New contributions should adhere to pep-8, but this is only weakly enforced
+New contributions should adhere to `PEP 8`_, but this is only weakly enforced
 (there is loads of legacy stuff that breaks it, and things like a single
 trailing whitespace is not a big deal).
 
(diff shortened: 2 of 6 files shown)


hgomersall:
Thanks for this. All excellent and necessary maintenance work.

I can't see `pyproject.toml`, is that somewhere else or did you forget to add it?

Also, I don't quite understand you point about the changes in numpy 1.10...?

grlee77:
> I can't see pyproject.toml, is that somewhere else or did you forget to add it?

Yeah, I forgot.  I have added it now.  

> Also, I don't quite understand you point about the changes in numpy 1.10...?

Regarding numpy 1.10, that seems to have been when casting of inputs to double precision as done here was added:
https://github.com/numpy/numpy/blob/b454ec7b9310608e24348aa03a09dccec04245fc/numpy/fft/fftpack.py#L283

In older numpy, passing long double to these routines would end up in an error being raised instead.  I assume that is the reason we were casting numpy inputs to double in the tests as done here:
https://github.com/pyFFTW/pyFFTW/blob/ad23f73defdb2ed27c834cadeff2fd772953a8a6/test/test_pyfftw_numpy_interface.py#L223-L227
That casting should not technically be necessary on recent numpy versions as it will be done automatically on numpy's end.  Not a big deal, it was just something I had noticed recently.

.travis.yml
```diff
@@ -9,9 +9,9 @@ env:
 
 python:
   - 2.7
-  - 3.3
   - 3.4
   - 3.5
+  - 3.6
```
jakirkham: 👍 Definitely time to drop 3.3. Nice to see 3.6.

README.rst
```diff
@@ -48,10 +48,10 @@ index page is `here <http://pypi.python.org/pypi/pyFFTW>`_.
 
 Requirements (i.e. what it was designed for)
 --------------------------------------------
-- Python 2.7 or greater (Python 3 is supported)
-- Numpy 1.6
+- Python 2.7 or greater (Python 3.4+ is also supported)
+- Numpy 1.10
```
jakirkham: It's probably worth keeping [this]( https://github.com/conda-forge/pyfftw-feedstock/blob/ad4b44b1bc636618c457ed3cb38d15fd4e046640/recipe/meta.yaml#L24-L26 ) in mind. We can obviously bump what we are building against. So this doesn't necessarily constitute a problem. Just depends on how much compatibility we are comfortable keeping.

README.rst
```diff
@@ -48,10 +48,10 @@ index page is `here <http://pypi.python.org/pypi/pyFFTW>`_.
...
 - FFTW 3.3 or higher (lower versions *may* work)
-- Cython 0.15 or higher (though the source release on PyPI loses this
+- Cython 0.23 or higher (though the source release on PyPI loses this
```
jakirkham: Just curious if there were any known issues this would be fixing. Certainly in favor of bumping Cython. Only wondering if there is anything else motivating it.

README.rst
grlee77: I would be fine with 1.8 as well.  There is no problem running with 1.6 at the moment, it is just that at some point it would be nice to be able to drop some of the workarounds for older numpy.  Perhaps use 1.8 for the next release and bump again to 1.10 at some future date?

jakirkham: Sure 1.8 seems fine.

Certainly not opposed to bumping more. It's worth noting that 1.10 wouldn't likely get lots of support in c-f, but 1.11 would. So it might make more sense to go to 1.11 or at least build with it in c-f. Assuming we want to go that way.

README.rst
grlee77: I just updated this text test to match what is currently be in our `setup.py`.  I am not sure exactly what the reason for 0.23 specifically was, but perhaps @fredRos may recall (I believe he added this version check in one of his recent PRs).  

https://github.com/pyFFTW/pyFFTW/blob/f4bdbfcad95b42fb1b0705c084d17f2d39d1d0a7/setup.py#L757

hgomersall:
I think we should bump everything since we're no longer testing against earlier versions, but I think we should make a distinction between _supported_ versions and _requirements_. Many of the bumps are not because we require the new version but because we don't verify support.

README.rst
jakirkham: As a counter to my original point, would add that Dask is tested with NumPy 1.10.4 at the lower end. Based on the `cbrt` wrapping issue seen here, we know it is not compatible with NumPy 1.9. So that might be another argument to move to NumPy 1.10 at a minimum. However it is an optional dependency.

grlee77:
I rebased and made one new commit where some text was added related to the requirements (and optional requirements).  I also added and cleaned up some links.  The `setuptools` documentation link was updated and the text related to`pkgin` on OS X was removed as the link was no longer functional.

TODO: 
   - [ ] There is still one mention of numpy 1.9 in the section on wheels.  We need to update this to match whatever version we are going to build the wheels with.

README.rst
```diff
@@ -38,26 +39,34 @@ Operating FFTW in multithreaded mode is supported.
...
 
 (install these as much as possible with your preferred package manager).
 
+In practice, pyFFTW *may* work with older versions of these dependencies, but
+it is not tested against them.
+
+Optional Dependencies
+---------------------
+- Scipy_ >= 0.16
+- Dask_ >= 0.14.2
```
jakirkham: Should be bumped to 0.15.0.

jakirkham: Should be resolved with PR ( https://github.com/grlee77/pyFFTW/pull/2 ).

jakirkham:
There were some conflicts. So put up PR ( https://github.com/grlee77/pyFFTW/pull/2 ) to resolve them as well as bump the Dask version requirement to 0.15.0.

jakirkham:
Should we include links to the conda-forge packages?

grlee77:
> Should we include links to the conda-forge packages?

I think it is probably worth adding.

grlee77:
> Should we include links to the conda-forge packages?

I think it is probably worth adding.

jakirkham:
FWIW there are conda/conda-forge badges on [Shields.io]( http://shields.io ), which would be an easy thing to add.

jakirkham:
Not sure how we want to lay them out, but here's [an example]( https://github.com/conda-forge/pyfftw-feedstock/blob/512782365bfcefe4f2322d5959d03ea182f43ef7/README.md#current-release-info ) of what they look like.

jakirkham:
LGTM. Thanks for doing this @grlee77.

jakirkham:
As to badges, put that in issue ( https://github.com/pyFFTW/pyFFTW/issues/220 ). It feels a bit out of scope for this PR even if they are interesting.

hgomersall:
Looks good to me. Does the trailing underscore in restructured text find the link for you?

grlee77:
> Does the trailing underscore in restructured text find the link for you?

It works that way: in sphinx  http://www.sphinx-doc.org/en/master/rest.html#external-links
and apparently for RST in general:  http://docutils.sourceforge.net/docs/user/rst/quickref.html#hyperlink-targets

FWIW, we have used that same style in PyWavelets and GitHub renders the README correctly

grlee77:
I think we are now good to merge here?

jakirkham:
Oops, accidentally pushed to your branch. Sorry about that. Well that resolves the merge conflict anyway.

jakirkham:
LGTM. Thanks @grlee77.

hgomersall:
Good stuff. Merge when ready (after the tests! :) )

pyproject.toml
```diff
@@ -0,0 +1,9 @@
+[build-system]
+requires = [
+    "wheel",
+    "setuptools",
+    "Cython>=0.23",
+    "numpy==1.10.4; python_version<='3.5'",
+    "numpy==1.12.1; python_version=='3.6'",
+    "numpy==1.13.1; python_version>='3.7'",
```
jakirkham: Thinking about this in the context of wheel builds being made available, should we be bumping the `numpy` requirement at run time based on what version we used at build time or does this somehow happen for us?

hgomersall: bleh! Wheels are a pain!

grlee77: I do not think this happens automatically, but I'm not sure if it is necessary in practice.  I have asked for some advice over at https://github.com/PyWavelets/pywt/pull/321#issuecomment-377542922

I [copied these particular values from scipy](https://github.com/pypa/pip/issues/4582#issuecomment-312111015), but bumped 1.9.3 to 1.10.4 to reflect our minimum NumPy requirement.  From what I can tell, SciPy still uses 1.8.2 in their `install_requires` in setup.py, but I'm not sure if there may be some other magic being performed elsewhere.

jakirkham: The concern would be if a wheel was built with a newer NumPy, but allowed to be installed on an older one where expected symbols are missing.

Looking at the pip thread linked in your pywt comment, I think they are relying on the fact that each version of Python comes online with a specific NumPy version to effectively impose this requirement. IOW NumPy 1.12 is the first version that supports Python 3.6. So if we build against NumPy 1.12 with Python 3.6, no older version of NumPy could be installed on Python 3.6. If I'm reading it correctly, then I think what you added here is correct. Happy to have someone else more involved with wheels confirm this though.

Yeah, that matches what the SciPy wheels show as well.

rgommers: > IOW NumPy 1.12 is the first version that supports Python 3.6. So if we build against NumPy 1.12 with Python 3.6, no older version of NumPy could be installed on Python 3.6.

That's correct, and this PR looks safe to me. However to make it more explicit, one could add an `install_requires={'numpy>=build_version'}`, where build version then needs to be determined at `setup.py` execution time based on the Python version.

hgomersall:
Much of the subtlety of this I'm not following, but I'm happy to defer to the wisdom of others. Should we merge this now?

grlee77:
I have removed `project.toml` from this PR.  I will make it a separate PR that can be merged later once issues in `pip` 10 have been resolved.

grlee77:
The Python 3.6 failures on AppVeyor is related to conda's numpy [using mkl_fft](https://ci.appveyor.com/project/hgomersall/pyfftw/build/1.0.454/job/q6o0w5ejyq3s160v?fullLog=true#L173) instead the native numpy.fft.  This apparently breaks a ton of the tests.
The Python 2.7 and 3.5 that passed did not install `mkl_fft`.

hgomersall:
Why is appveyor failing?

grlee77:
It looks like the failures on Python 3.6 may be because that is the only version `mkl-fft` is being installed on.  I am trying again now with `nomkl` specified for the conda environment.

jakirkham:
`defaults` has also been exploring a `blas` package similar to `conda-forge`’s. So installing `blas=*=openblas` may work.

grlee77:
Thanks @jakirkham. Apparently `nomkl` didn't prevent various MKL packages from being installed alongside numpy. I tried making a new environment with`blas=*=openblas` locally and at least on linux it did not pull in any MKL stuff. We will find out now if it works on Win/Appveyor.

grlee77:
Well.  `nomkl` isn't available on windows and `'blas=*=openblas` apparently doesn't work there either`.  I am going to try using `conda-forge` just for Python 3.6 as a workaround unless there are better ideas?  I suppose with wheels being available for all of these nowadays we could always just use pip instead?

grlee77:
I rebased all the Appveyor testing commits into ebb0e00 once I found an approach that worked.  I have also added a new commit that removes the numpy version check as recently discussed in #231. please review.

If you prefer a different solution such as switching to pip for Appveyor, let's address that in a separate PR.

.travis.yml
jakirkham: At some point, we should add Python 3.7, but that is currently blocked on issue ( https://github.com/travis-ci/travis-ci/issues/9815 ).

README.rst
jakirkham: Guess I forgot to update this in the text as well. Would you like to update or would you like me to send another PR?

appveyor.yml
```diff
@@ -30,11 +33,11 @@ environment:
           PYTHON_ARCH: "64"
           PYTHON_VERSION: "2.7"
 
-        - PYTHON_HOME: "C:\\Miniconda3"
+        - PYTHON_HOME: "C:\\Miniconda34"
           PYTHON_ARCH: "32"
           PYTHON_VERSION: "3.4"
 
-        - PYTHON_HOME: "C:\\Miniconda3-x64"
+        - PYTHON_HOME: "C:\\Miniconda34-x64"
```
jakirkham: Side note: We can just use the latest Miniconda as the Python version used by `conda` doesn't affect the Python version we use in our testing environment.

jakirkham:
Looks pretty good to me. Couple minor comments above.

hgomersall:
Marvellous stuff :+1:

Issue #205, opened 2018-01-23
monkey patching issue


dvolgyes:
I think the monkey patching example became a bit outdated.
I mean it works well, except the speedup.
I think the reason is this:
scipy.signal.fftconvolve() is used in the example, but the inner part of fftconvolve
has changed, and now it uses rfftn and irfftn from scipy.fft_pack.
However, these two are not implemented in the pyFFTW interface.
(fftn and ifftn are implemented, also rfft and irfft, but not the rfftn and the irfftn.)

Due to the lack of implementation, I assume the code falls back to the scipy
version, and therefore there is no speedup.

I would recommend to implement these two missing interfaces,
or to change the monkey patching example.

hgomersall:
Oh, I hadn't noticed that - thanks! Fancy submitting a PR?

dvolgyes:
In general, yes, but at this moment
I am very buried in deadlines until April.
If nobody sends PR until April, then I will try to make one.

grlee77:
I took a quick look at this and there is not an `rfftn` or `irfftn` in scipy's fftpack, so I don't think we are missing anything there.  

You are correct that fftconvolve uses rfftn/irfftn when the input data is real-valued, but it [uses the ones from numpy](https://github.com/scipy/scipy/blob/beb258ef9f9b76567f094984f5bfeed774e7e9eb/scipy/signal/signaltools.py#L391-L393) which we do have in our numpy interfaces.

For complex-valued input as in the [example from the docs](http://pyfftw.readthedocs.io/en/latest/source/tutorial.html#monkey-patching-3rd-party-libraries), it should be scipy.fftpack.fftn/scipy.fftpack.ifftn that are used, so I think the example is okay as is.  

We could expand the example to show the correct usage for real-valued data.  I guess one would have to monkeypatch `numpy.fft` to get the speedup in `scipy.fftpack` on real-valued data which is not obvious without inspecting the internals of fftconvolve!  

Also, along these lines, the default numpy, scipy and builder interfaces all use a single thread which is likely not optimal on many modern systems.  I had considered implementing the ability to set an environment variable such as `PYFFTW_NUM_THREADS` that would override this default.  I implemented this recently, but never got around to making a PR for it.  Does that sound like something that would be of interest?  I think the only way to do this now this now would be for the user to roll their own `functools.partial` wrappers around the existing interfaces and monkey patch using those, which is a bit of a pain.

grlee77:
[...]

PR #208, opened 2018-01-30
configure the number of threads via an environment variable

grlee77:
Currently the interfaces and planners default to a single thread.  This makes it a pain to monkey-patch numpy or scipy with multi-threading enabled (as discussed in #205).

This PR causes pyFFTW to check the environment variable `PYFFTW_NUM_THREADS` upon import and use its value as the default number of threads in all planners and interfaces.  If the variable has not been defined, the current default of 1 thread is used.  If `PYFFTW_NUM_THREADS` is <= 0 then multiprocessing.cpu_count() is used to select all available threads.

@hgomersall, in #205, you had suggested some sort of lookup, but I wasn't clear on what you meant.  Can you think of a better approach to let the user configure this?

**TODO:**
- [ ] documentation
- [ ] test case(s)
- [ ] always set the number of threads to 1 if pyFFTW wasn't built with pthreads or OpenMP support
- [ ] extend to allowing the user to select the default planner effort as well?

### DIFF
--- a/pyfftw/__init__.py
+++ b/pyfftw/__init__.py
@@ -15,10 +15,10 @@ import os
 
 # All planners and interfaces default to single threaded unless otherwise
 # specified via PYFFTW_NUM_THREADS.
-default_num_threads = int(os.environ.get('PYFFTW_NUM_THREADS', 1))
-if default_num_threads <= 0:
+_default_num_threads = int(os.environ.get('PYFFTW_NUM_THREADS', 1))
+if _default_num_threads <= 0:
     import multiprocessing
-    default_num_threads = multiprocessing.cpu_count()
+    _default_num_threads = multiprocessing.cpu_count()
 
 from .pyfftw import (
         FFTW,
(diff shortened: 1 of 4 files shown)


hgomersall:
@grlee77 I meant: is it possible to infer the number of cores from some system information?

hgomersall:
Yeah, this seems reasonable. The proposed solution is good.

grlee77:
I think this is generally good to go.  If you set the environment variable prior to starting the python console it works well.

One drawback is that if you have already imported pyFFTW and then try to set PYFFTW_NUM_THREADS (e.g. via `os.environ`) it is not going to have any effect in that session.  I don't think that is a major limitation, but if run-time configuration is desired we need an alternative solution.

>a job may be dispatched to a  node that has X resources, but the job is only given Y. Depending on how this is implemented, info from the node resources can leak through.

This is one argument in favor of leaving the default number of threads as 1 (I did it here primarily for backward compatibility).  It also avoids potential conflict with other sources of multithreading such as dask on a user's machine.  Users that need multithreading from the pyFFTW interfaces can enable it fairly painlessly via the environment variable.

fredRos:
If pyfftw is build with openMP, you can just set `OMP_NUM_THREADS` or use any other other runtime functions to set the number of threads. with pthreads, it's not easy, unfortunately

hgomersall:
@fredRos

PR #210, opened 2018-03-12
setup.py: make sure install_requires contains numpy

grlee77:
We need to make sure `install_requires` always contains numpy.  This was not currently always the case which I suspect is the underlying reason the wheels built in conda-forge/pyfftw-feedstock#16 did not cause pip to install a numpy wheel.  

Also, was Cython really needed in `install_requires`?  I have tried removing it here.

cc @jakirkham

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -742,10 +742,11 @@ def setup_package():
 
     # Figure out whether to add ``*_requires = ['numpy']``.
     build_requires = []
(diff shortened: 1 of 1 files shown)


jakirkham:
Just took this for a spin and it worked as intended. Thanks @grlee77. :)

PR #211, opened 2018-03-12
Fix Dask test interface

jakirkham:
Appears there was some copy-pasta in the tests that caused the NumPy FFT interface to be used for some things. This corrects it to be the Dask FFT interface instead.

As the test infrastructure uses some magic to find default arguments and use them in the tests, it gets confused by how the Dask FFT wrapper works. So we use the NumPy FFT interface (what Dask wraps) for the inference magic to inspect.

Due to these changes, a few argument tests that are not valid with the Dask interface fail. As the interface doesn't currently allow arbitrary arguments to be passed through or may deviate in terms of what exceptions it raises, we just have to drop these test cases as they simply are not supported at present. Raised issue ( https://github.com/dask/dask/issues/3278 ) upstream about getting arbitrary keyword arguments passed through. We can also discuss adjustments to the exceptions raised upstream if there is interest.

### DIFF
--- a/pyfftw/interfaces/__init__.py
+++ b/pyfftw/interfaces/__init__.py
@@ -237,14 +237,9 @@ else:
     from . import scipy_fftpack
 
 try:
-    import dask.array.fft
-    dask.array.fft.fft_wrap
-except (ImportError, AttributeError):
+    from dask.array.fft import fft_wrap
+except ImportError:
     pass
 else:
+    del fft_wrap
     from . import dask_fft
-
-try:
-    del dask
-except NameError:
-    pass
--- a/test/test_pyfftw_dask_interface.py
+++ b/test/test_pyfftw_dask_interface.py
@@ -46,7 +46,7 @@ try:
     import dask.array as da
     from dask.array import fft as da_fft
     from dask.array.fft import fft_wrap
-except (ImportError, AttributeError):
+except ImportError:
     da = None
     da_fft = None
 
--- a/test/test_pyfftw_dask_interface.py
+++ b/test/test_pyfftw_dask_interface.py
@@ -108,7 +108,8 @@ class InterfacesDaskFFTTestFFT(unittest.TestCase):
             'c2r': (complex_dtypes, make_complex_data)}
 
     validator_module = da_fft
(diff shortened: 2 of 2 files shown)


grlee77:
The changes look good.  Is it worth skipping the failing tests rather than deleting them? (so they could be easily enabled later if `kwargs` support is adopted upstream)

jakirkham:
Honestly they were just copied from `numpy_fft`. Figure we can copy them again should we start supporting those options.

That said, some options like threads don't make lots of sense in the context of Dask where it is already using some form of parallelism outside of what pyFFTW/FFTW offer. So IDK if we want to go the route of supporting these yet.

jakirkham:
The CI failure is a little strange. Last time we played with this we found the Dask imports at the top of this test file were sufficient to determine if we needed to skip testing the Dask interface. Now it complains `dask_fft` is not found in `interfaces`. So updated the check to verify exactly that `dask_fft` is in `interfaces`, yet it still fails claiming it wasn't found. Not sure how both can be true. 😕

grlee77:
That is odd (also strange that it only occurred on 3.3 and 3.4).

Possibly related: I did slightly change the `dask` import logic in `__init__.py` a little while back:  fbad1bde61a31eb91c6f77520807fef327354f96

The change was due to a case where a missing function in our numpy wrappers was causing dask itself to raise an AttributeError during import.  Then the `del dask` line that was under the AttributeError case became invalid because dask had never been imported.  (We have since added the missing `rfftfreq` function to the numpy wrappers as well)

jakirkham:
Ah, sorry, that's my fault. I'll take a look at that code and see if we can't simplify a bit.

jakirkham:
Ok, think that straightens out the CI issues. Worked in a Python 3.4 environment locally. Morale of the story is code defined within the class definition will be run whether or not that test case is skipped. Added a workaround to fix the issue, which allows the tests to be skipped. Did a bit of simplifying in the conditional import logic. Hopefully that's a bit cleaner and more reliable. Please let me know what you think.

jakirkham:
Thanks all. :)

PR #214, opened 2018-03-13
Handle all exceptions importing Dask causes

jakirkham:
If importing Dask triggers any exception whatsover, simply catch that exception. In these cases, assume that Dask is not available and proceed without it.

### DIFF
--- a/pyfftw/interfaces/__init__.py
+++ b/pyfftw/interfaces/__init__.py
@@ -238,7 +238,7 @@ else:
 
 try:
     from dask.array.fft import fft_wrap
-except ImportError:
+except Exception:
     pass
 else:
     del fft_wrap
--- a/test/test_pyfftw_dask_interface.py
+++ b/test/test_pyfftw_dask_interface.py
@@ -51,7 +51,7 @@ try:
     import dask.array as da
     from dask.array import fft as da_fft
     from dask.array.fft import fft_wrap
-except ImportError:
+except Exception:
     da = None
     da_fft = None
 


jakirkham:
Saw the old AppVeyor `AttributeError` failure come back, which I think I introduced. Sorry about that. Might help to have the AppVeyor status shown ( https://github.com/pyFFTW/pyFFTW/issues/215 ) to avoid that. In any event, expect this should fix the issue.

hgomersall:
I'm not so keen on the catch on Exception. Can we make it explicit?

grlee77:
let's just go back to using `(ImportError, AttributeError)` instead of `Exception`.

jakirkham:
To provide some context, Dask wraps up some ufuncs from NumPy. One of them is `cbrt`. However `cbrt` is only added in NumPy 1.10. It appears we are installing NumPy 1.9 on AppVeyor (though it doesn't appear to be intentional) hence the `AttributeError`. Ideally Dask should include a lower bound on NumPy. Added PR ( https://github.com/dask/dask/pull/3283 ) upstream to address this. Though the version of Dask on AppVeyor is also quite old (not sure why 🤔).

Given that Dask currently does not have a lower bound, am a little concerned what might happen if a user mixes NumPy 1.6 and Dask for instance. Not sure if they were ever compatible, but wouldn't be surprised to see some other strange exceptions get generated. Would like to just catch them altogether as opposed to catching different possible errors in a piecemeal fashion to avoid reports like these later. Admittedly using `Exception` is not great.

More generally, we may be overly concerned here if we plan on just dropping NumPy 1.9 altogether.

hgomersall:
1.9 is pretty old. Can we just say we don't support it and handle things properly? I mean, this is what supporting means right?

jakirkham:
To be clear, I'm trying to say this is a problem with Dask's version constraints not pyFFTW. Namely Dask shouldn't be installable when NumPy 1.9 or earlier is present. That said, we still have to do something in the interim.

Fortunately we have a few options (listed below). If you can think of others, please feel free to share.

1. Require NumPy 1.10+
2. Fix CIs to only install Dask on NumPy 1.10+
3. Check NumPy is 1.10+ before adding Dask interface
4. Do some `try...except` workaround like this one or the previous one.

To be honest, I'm starting to lean towards anything in 1-3. Slight preference towards 3. Thoughts?

jakirkham:
Implementing 3 instead in PR ( https://github.com/pyFFTW/pyFFTW/pull/218 ).

PR #217, opened 2018-03-13
Handle unspecified NumPy version

jakirkham:
If the NumPy version is not specified, drop the `=`.


jakirkham:
Was trying to see if this was causing an old version of NumPy to get installed accidentally. Also wanted to see if this might resolve the issues with testing Dask against unsupported NumPy versions. If it doesn't work, was going to try conditionally installing Dask based on a NumPy version check.

PR #218, opened 2018-03-13
Explicitly add Dask Array as an extra requirement

jakirkham:
Lists Dask Array as an extra requirement of pyFFTW. Make sure this requirement pulls in Dask with the Array extras and requires NumPy 1.10+. Also note this in the `requirements.txt` file as well.

In the event that Dask and pyFFTW end up in an environment without a compatible NumPy (e.g. older than NumPy 1.10), add a guard against adding the Dask interface that checks the NumPy version.

### DIFF
--- a/pyfftw/interfaces/__init__.py
+++ b/pyfftw/interfaces/__init__.py
@@ -236,10 +236,18 @@ else:
     del scipy
     from . import scipy_fftpack
 
-try:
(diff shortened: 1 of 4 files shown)


jakirkham:
Thought we could use the same thing for SciPy. So put together PR ( https://github.com/pyFFTW/pyFFTW/pull/219 ) with that change.

PR #221, opened 2018-03-16
Optionally use Dask if Numpy is 1.10

jakirkham:
The previous Dask interface import guard check required NumPy to be greater than 1.10.0 to work. However it is also ok for NumPy to be equal to 1.10.0. This relaxes the check a bit to allow NumPy 1.10.0 for instance. However filters out dev and rc releases. The latter should have no practical importance given how old NumPy 1.10 is. Nevertheless it is good to be strict about having a formal release to avoid potential issues.

### DIFF
--- a/pyfftw/interfaces/__init__.py
+++ b/pyfftw/interfaces/__init__.py
@@ -236,16 +236,16 @@ else:
     del scipy
     from . import scipy_fftpack
 
-from distutils.version import LooseVersion
+from distutils.version import StrictVersion
 import numpy
 
 fft_wrap = None
-if LooseVersion(numpy.__version__) > LooseVersion("1.10"):
+if StrictVersion(numpy.__version__) >= StrictVersion("1.10.0"):
     try:
         from dask.array.fft import fft_wrap
     except ImportError:
         pass
-del LooseVersion
+del StrictVersion
 del numpy
 
 if fft_wrap:


jakirkham:
@hgomersall, any thoughts on this?

jakirkham:
FWIW this may be irrelevant in light of PR ( https://github.com/pyFFTW/pyFFTW/pull/204 )'s `numpy` requirements. That said, would still be nice to have this in for the interim (especially if we decide to support older `numpy`'s longer).

jakirkham:
Is this ok to go in? Questions? Comments? Concerns?

pyfftw/interfaces/__init__.py
```diff
@@ -236,16 +236,16 @@
     del scipy
     from . import scipy_fftpack
 
-from distutils.version import LooseVersion
+from distutils.version import StrictVersion
 import numpy
 
 fft_wrap = None
-if LooseVersion(numpy.__version__) > LooseVersion("1.10"):
+if StrictVersion(numpy.__version__) >= StrictVersion("1.10.0"):
```
grlee77: I was going to suggest using
```python
if np.lib.NumpyVersion(np.__version__) >= '1.10.0':
```

to avoid the distutils import, but apparently that function was only added in numpy 1.9.  I think what you have is preferable in order to not break the import on older numpy.

jakirkham: Interesting, did not know about that.

jakirkham:
Thanks for the review @grlee77. :)

PR #224, opened 2018-04-17
WIP: Fix windows wheel upload to bintray

matmel:
Trying to fix the problem (and potential fix) mentioned in issue comment https://github.com/pyFFTW/pyFFTW/issues/187#issuecomment-382014237

### DIFF
--- a/appveyor/deploy.ps1
+++ b/appveyor/deploy.ps1
@@ -2,6 +2,7 @@
 function deploy_to_bintray() {
     $version_string = (iex "python -c `"import pyfftw; print(pyfftw.__version__)`"") | Out-String
     $version = $version_string -replace "`t|`n|`r",""
+    $short_version = [string]$version.Split("+")[0]
 
     if ($env:PYTHON_ARCH -eq "32") {
         $platform_suffix = "win32"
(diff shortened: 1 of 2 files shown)


matmel:
So the failure here happened during the python 3.4 build where conda command was not found. I suppose there is something wrong on appveyor VMs. Is it still needed to build `pyFFTW` for python 3.4 though?

More importantly, my modifications were not exercised at all. If you compare last commit on master before mine: the deployment phase looks like [this](https://ci.appveyor.com/project/hgomersall/pyfftw/build/1.0.450/job/rvs5udwr80lhclmt:L1157), but my commit build does not pass through that phase: see [here](https://ci.appveyor.com/project/hgomersall/pyfftw/build/1.0.451/job/pa5jqyemlrtl9olj:L1160)

I'm not familiar enough with appveyor but is it related to some authorization given to some people only (which wouldn't be surprising).

hgomersall:
I'm happy to drop 3.4 for Windows.

appveyor/deploy.ps1
```diff
@@ -17,8 +15,8 @@ function deploy_to_bintray() {
     $username_password = "${env:bintray_username}:${env:bintray_api_key}"
 
     Write-Host "Uploading: $filepath"
-    Write-Host "to https://api.bintray.com/content/hgomersall/generic/PyFFTW-development-builds/$short_version/$filename"
```
hgomersall: I think we still need a short version (e.g. 0.10.0) for the path, otherwise every dev release will produce a new path. This can be extracted from the version string quite easily I think (split on + for dev releases). What do you think? Happy to change my mind on this.

hgomersall:
I'm happy for you to remove 3.4 from the matrix if it works :)

matmel:
OK, I removed python 3.4 from the build matrix.

Do you know why the `appveyor/deploy.ps1` script is not called in appveyor jobs?

hgomersall:
@matmel I suspect it's an appveyor policy, otherwise every PR would be deployed!

hgomersall:
@grlee77 I'm happy with removing 3.4 from the windows test matrix with this, but I'd like a second opinion if you have a moment.

grlee77:
removing 3.4 is fine with me as well

grlee77:
The small change in 1a435eb fixes Python 3.4 on Appveyor if you would like to keep it.

hgomersall:
@matmel if we revert the 3.4 stuff and use @grlee77's fix, that seems slightly better to me.

grlee77:
I made a copy of this in #248 that removes the final commit. Thanks @matmel (you are still credited in #248).

PR #226, opened 2018-04-25
add pyproject.toml

grlee77:
Addition of `pyproject.toml` was removed from #204 due to current compatibility issue with pip 10.0.0. This PR should not be merged until after these issues have been resolved.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,7 +47,7 @@ before_install:
 install:
   - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION setuptools cython $NUMPY_SPEC $SCIPY_SPEC $DASK_SPEC blas=*=openblas
   - conda activate test-environment
-  - python setup.py -v build_ext --inplace
+  - pip install -e . -v --no-build-isolation
 
 script:
   - python setup.py test
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -7,6 +7,7 @@ include pyfftw/pyfftw.pxd
 include pyfftw/cpu.pxd
 include pyfftw/utils.pxi
 include test/*.py
+include pyproject.toml
 recursive-include include *.h
 
 # All documentation
--- a/README.md
+++ b/README.md
@@ -73,12 +73,12 @@ package index with tools such as pip:
 
     pip install pyfftw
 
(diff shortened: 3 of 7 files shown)


grlee77:
I am going to close this for now since pip is still not ready.  We can reopen at a later date

grlee77:
This should work as of pip 18.0. The selection of a NumPy version based on the Python version as specified here did not work properly when `pyproject.toml` was first introduced in pip 10.0, so I had closed this PR previously.

For more info on `pyproject.toml` see [PEP-518](https://www.python.org/dev/peps/pep-0518/) and specifics on its [implementation in pip](https://pip.pypa.io/en/stable/reference/pip/#pep-518-support).  Specifically in pip's implementation, the presence of `pyproject.toml` triggers build isolation so that the build dependencies do not get installed into the user's runtime environment.

To test this locally, I created a new python 3.5 environment and then checked out this branch from GitHub. Running `pip install . -v` from within the source directory resulted in the expected version for NumPy and other dependencies as indicated by the following line during the pip install:

`Successfully installed Cython-0.29.2 numpy-1.10.4 setuptools-40.6.3 wheel-0.32.3`

After the build completed, my environment had NumPy 1.15.4, but no Cython (NumPy 1.10.4 and Cython 0.29.2 were installed into the isolated build environment by pip and did not end up in my runtime environment).

grlee77:
Based on [more recent pip docs](https://pip.pypa.io/en/latest/reference/pip/#pep-517-and-518-support), it looks like one can also explicitly specify setuptools as the build backend by adding the following line:
 `build-backend = "setuptools.build_meta"`
but it seems that this is the default setting if it is not explicitly specified.
Explicitly specifying the build backend is related to  [PEP-517](https://www.python.org/dev/peps/pep-0517/)

hgomersall:
I like this solution! Does it require pip>=18 to be already installed?

grlee77:
I updated the `pyproject.toml` file. Compatibility with `versioneer` required a small change in 5202a37 so that the local copy of versioneer gets found in the isolated build environment.

I also updated `requirements.txt` and the versions listed in `setup.py` which were overlooked in #301 (which updated them in the CI scripts and README)

hgomersall (approved):
Looks good to me!

Issue #229, opened 2018-05-05
License is misleading; programs using pyFFTW still need to comply with the GPL


brakhane:
pyFFTW claims to be licensed under 3-clause BSD, however, for pyFFTW to do anything, it needs to use the FFTW library, which is GPL licensed. Therefore, [any program using it must also be GPL licensed](https://www.gnu.org/licenses/gpl-faq.en.html#GPLWrapper), unless they buy a license from MIT.

To avoid people unintentionally violating the FFTW license, I suggest to add a clarification at the beginning of the license and in the documentation that every program using it must effectively be released with the GPL license (unless a license is purchased from MIT)

hgomersall:
This is actually a very grey area when it comes to Python, but I'm actually happy with the principle. A pull request as a clarification to LICENSE would be appreciated.

brakhane:
> This is actually a very grey area when it comes to Python

not sure what it has to do with Python and why it's a grey area.

Created a PR

mreineck:
I think this is a misinterpretation of the linked FAQ.

pyFFTW can be licensed under BSD terms, and any code using pyFFTW can be licensed in that way as well. ONLY if you distribute a "combined package" (e.g. in form of a binary distribution), the whole thing must be covered by GPL terms. But even then nobody is stopping you from distributing the pyFFTW sources additionally under a BSD license.

hgomersall:
I'm confused as the the confusion...

mreineck:
Sorry ... I just encountered this issue while looking for other pyfftw-related information.

I just wanted to clarify the original statement:

> Therefore, any program using it must also be GPL licensed, unless they buy a license from MIT.

To the best of my knowledge this is not correct. The correct statement would be:

"Therefore, any program using it must also have a GPL-commpatible license, unless they buy a license from MIT."

And BSD is GPL compatible.

hgomersall:
I'm happy with that - it's the mechanism under with pyfftw is licensed. Are you able to raise such a PR?

brakhane:
According to [the GPL faw](https://www.gnu.org/licenses/gpl-faq.html#WhatDoesCompatMean), "compatible" means that you can combine a compatible licensed program with a GPL program, *however* if you distribute the resulting program, you have to license it using the (same) GPL license.

hgomersall:
Right, but it's possible to distribute the code that is not linked to the GPL code under a more liberal license. "compatible" means you can re-license under the GPL as necessary.

[...]

PR #230, opened 2018-05-07
Add clarification to license and add license to docs.

brakhane:
Code that uses FFTW3 must either be licensed under the GPL,
or the user must buy a commercial license from MIT.

Add a note to the license that while pyfftw is licensed under
BSD, it de facto cannot be used in proprietary programs without
acquiring a license from MIT.


brakhane:
Related to #229

PR #231, opened 2018-07-16
less strict Numpy version comparison on import

grlee77:
I happened to be working in a conda environment with a development build of numpy:
```numpy:           1.15dev-py36_blas_openblas_201 local       [blas_openblas]```

In this environment, the value of `numpy.__version__` is `1.15.0.dev0+7bb2d5a`. This version string leads to the following error in StrictVersion during import of pyFFTW:

```ValueError: invalid version number '1.15.0.dev0+7bb2d5a'```

Using either `distutils.version.LooseVersion` or `numpy.lib.NumpyVersion` seems to resolve the issue.

### DIFF
--- a/pyfftw/interfaces/__init__.py
+++ b/pyfftw/interfaces/__init__.py
@@ -236,7 +236,12 @@ else:
     del scipy
     from . import scipy_fftpack
 
-from numpy.lib import NumpyVersion
+try:
+    # NumpyVersion was introduced in NumPy 1.9
+    from numpy.lib import NumpyVersion
+except ImportError:
+    from distutils.version import LooseVersion as NumpyVersion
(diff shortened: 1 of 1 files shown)


jakirkham:
IIRC we were going to bump the minimum NumPy version to 1.10. Is that right? If so, maybe we can just do that and drop this check. Thoughts?

grlee77:
> IIRC we were going to bump the minimum NumPy version to 1.10. Is that right? 

Yes, that is what is in #204.  I had forgotten that was still pending.

hgomersall:
Ah crap, I've been out of the loop for a while - been on vacation and also super busy. Where are we? Does #204 solve this problem?

PR #239, opened 2018-09-04
Try to clarify FFTW licensing conditions

mreineck:


brakhane:
As discussed in #229, I am convinced this clarification is wrong.

PR #240, opened 2018-09-12
avoid multi-indexing warnings with numpy >= 1.15

grlee77:
Indexing with lists of slices is raises a `FutureWarning` in numpy 1.15. This was occurring in the scipy interfaces and in some of the tests leading to tons of warnings in the test suite. This PR just converts these lists of slices to tuples.

PR #241, opened 2018-09-12
Allow run-time choice of the default number of threads and planning effort

grlee77:
This PR supersedes #208 (I think it is a better solution as it allows setting these at run-time as well). I am not sure why #208 was closed last time around, but I think some ability to configure the number of threads (and planner) used by the builders and interfaces is important.

Currently the interfaces and planners default to a single thread. This makes it a pain to monkey-patch numpy or scipy with multi-threading enabled (as discussed in #205).

With the solution in this PR, defaults for the number of threads are read at import time from `PYFFTW_NUM_THREADS` if it is defined (or if OpenMP-based threading is being used there is a fallback to `OMP_NUM_THREADS` when `PYFFTW_NUM_THREADS` is undefined). If neither environment variable is defined, the current default of 1 thread is used.

Similarly the environment variable `PYFFTW_PLANNER_EFFORT` can be used to select a different default planner effort. If it is undefined, the current default of `PYFFTW_ESTIMATE` is used.

Unlike in #208, the user can also modify these at runtime as follows:
```
from pyfftw import config

config.NUM_THREADS = 8
config.PLANNER_EFFORT = 'FFTW_MEASURE'
```
Subsequent calls to the builders or interfaces will then use these values instead. 

The implementation in `config.py` is a simplified version of Numba's config.py (their version also allows checking against a .yaml configuration file, etc., but I didn't think that was necessary in this case).

Please let me know if this implementation looks acceptable and I will add a commit to update the documentation.

**TODO:**
- [x] add documentation

### DIFF
--- a/docs/source/pyfftw/pyfftw.rst
+++ b/docs/source/pyfftw/pyfftw.rst
@@ -98,3 +98,35 @@ Utility Functions
 .. autofunction:: pyfftw.is_n_byte_aligned
 
 .. autofunction:: pyfftw.next_fast_len
+
+.. _configuration_variables:
+
+FFTW Configuration
+------------------
+
+.. data:: pyfftw.config.NUM_THREADS
+
+   This variable controls the default number of threads used by the functions
+   in :mod:`pywt.builders` and :mod:`pywt.interfaces`.
+
+   The default value is read from the environment variable
+   ``PYFFTW_NUM_THREADS``. If this variable is undefined and the user's
+   underlying FFTW library was built using OpenMP threading, the number of
+   threads will be read from the environment variable ``OMP_NUM_THREADS``
+   instead. If neither environment variable is defined, the default value is 1.
+
+   If the specified value is ``<= 0``, the library will use
+   :func:`multiprocessing.cpu_count` to determine the number of threads.
+
+   The user can modify the value at run time by assigning to this variable.
+
+.. data:: pyfftw.config.PLANNER_EFFORT
+
+   This variable controls the default planning effort used by the functions
+   in :mod:`pywt.builders` and :mod:`pywt.interfaces`.
+
+   The default value of is determined by reading from the environment variable
+   ``PYFFTW_PLANNER_EFFORT``. If this environment variable is undefined, it
+   defaults to ``'FFTW_ESTIMATE'``.
+
+   The user can modify the value at run time by assigning to this variable.
--- a/docs/source/pyfftw/pyfftw.rst
+++ b/docs/source/pyfftw/pyfftw.rst
@@ -107,7 +107,7 @@ FFTW Configuration
 .. data:: pyfftw.config.NUM_THREADS
 
    This variable controls the default number of threads used by the functions
-   in :mod:`pywt.builders` and :mod:`pywt.interfaces`.
+   in :mod:`pyfftw.builders` and :mod:`pyfftw.interfaces`.
 
    The default value is read from the environment variable
    ``PYFFTW_NUM_THREADS``. If this variable is undefined and the user's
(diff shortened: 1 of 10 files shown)


pyfftw/config.py
```diff
@@ -0,0 +1,82 @@
...
+        self.update(force=True)
+
+    def update(self, force=False):
+        new_environ = {}
+
+        # read local env var OMP_NUM_THREADS and any starting with PYFFTW_
+        for name, value in os.environ.items():
+            if name.startswith('PYFFTW_') or name == 'OMP_NUM_THREADS':
+                new_environ[name] = value
+        # We update the config variables if at least one NUMBA environment
```
hgomersall: Presumably this is not NUMBA any more? ;)

hgomersall:
Appveyor is broken. Does #224 fix this?

grlee77:
The Appveyor issue is related to it being unable to find a scipy that works with the OpenBLAS version of numpy.  Anaconda on some python/numpy versions bundles a numpy that uses `mkl_fft` in place of numpy's `fftpack`, but `mkl_fft` has a few issues that cause test failures. To avoid this, I was trying in #240 to switch to installing numpy and dependent packages from pip instead so that a non-MKL version will be found.

grlee77:
I don't think #224 is related to the issue seen here

grlee77:
#242 fixes the Appveyor failure that was unrelated to the content of this PR.

grlee77:
@hgommersall: API & tutorial documentation was now added if you want to take a quick look.

docs/source/pyfftw/pyfftw.rst
```diff
@@ -98,3 +98,35 @@ Utility Functions
...
+
+.. _configuration_variables:
+
+FFTW Configuration
+------------------
+
+.. data:: pyfftw.config.NUM_THREADS
+
+   This variable controls the default number of threads used by the functions
+   in :mod:`pywt.builders` and :mod:`pywt.interfaces`.
```
hgomersall: should this not be `pyfftw`?

docs/source/pyfftw/pyfftw.rst
```diff
@@ -98,3 +98,35 @@ Utility Functions
...
+
+   If the specified value is ``<= 0``, the library will use
+   :func:`multiprocessing.cpu_count` to determine the number of threads.
+
+   The user can modify the value at run time by assigning to this variable.
+
+.. data:: pyfftw.config.PLANNER_EFFORT
+
+   This variable controls the default planning effort used by the functions
+   in :mod:`pywt.builders` and :mod:`pywt.interfaces`.
```
hgomersall: also here

docs/source/tutorial.rst
```diff
@@ -102,14 +102,19 @@ Monkey patching 3rd party libraries
 Since :mod:`pyfftw.interfaces.numpy_fft` and
 :mod:`pyfftw.interfaces.scipy_fftpack` are drop-in replacements for their
 :mod:`numpy.fft` and :mod:`scipy.fftpack` libraries respectively, it is
-possible use them as replacements at run-time through monkey patching.
+possible use them as replacements at run-time through monkey patching. Note
```
hgomersall: possible _to_ use them?

docs/source/pyfftw/pyfftw.rst
grlee77: yes, definitely.  thanks!

grlee77:
also, as noted in #205 the example of monkeypatching using `fftconvolve` doesn't really show any substantial speedup for me. If I change the example to just do a basic fft such as using `scipy.fftpack.fftn` there is a speedup which becomes quite large in the case where the shape is unfavorable to fftpack. It may be worth updating that example in a separate PR.

hgomersall:
@grlee77 Yeah, it's been years since I looked at that! :)

hgomersall:
I approve this pull request :+1:

jakirkham:
Neat! Thanks for working on this @grlee77 :)

PR #242, opened 2018-09-15
update test suite avoid mkl_fft when it is present

grlee77:
seeing if this workaround gets the tests to pass with Numpy + mkl_fft. These were previously failing on Appveyor. 
In the past we tried to force OpenBLAS-based numpy on AppVeyor, but there seems to be a lack of Scipy + OpenBLAS packages on conda defaults and conda-forge. I'm not sure how it previously worked, but no longer does. Here I try a different strategy of using functions from within numpy.fft.fftpack instead of numpy.fft, hoping that this will avoid use of mkl_fft when it is present.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -56,7 +56,7 @@ environment:
           NP_BUILD_DEP: "1.14"
           SP_BUILD_DEP: "1.1"
           CYTHON_BUILD_DEP: "0.28"
-          EXTRA_CONDA_ARGS: "-c conda-forge blas=*=openblas"  # needed for non-MKL numpy/scipy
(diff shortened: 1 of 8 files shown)


grlee77:
The first two commits just use `numpy.fft.fftpack.fft` instead of `numpy.fft.fft`, etc when `mkl_fft` is present. This seems to have fixed all test failures except for those related to the dask interface tests (where mkl_fft is being used indirectly within dask itself). I added a final commit that just disables the dask interface tests when `mkl_fft` is also present. 

These changes allow everything to pass on Appveyor without having to try and restrict to OpenBLAS numpy + scipy (which is what was causing recent Appveyor failures due to a lack of OpenBLAS-based scipy).

PR #250, opened 2018-11-17
prevent absolute path from ending up in the egg-info SOURCES.txt

### LINKED ISSUE #249: absolute path ending up in SOURCES.txt on PyPI
@jakirkham, @hgomersall:
I just encountered an odd issue with the current release. 

There is an absolute path getting stored within the `pyFFTW.egg-info/SOURCES.txt` file included with the source .tar.gz on PyPI. I am not yet sure what is causing this to appear there, but its presence is currently causing the conda-forge build to fail:

https://ci.appveyor.com/project/conda-forge/pyfftw-feedstock/build/job/jxmffjos8pc21sik#L855

I originally generated the source tarball by running the following:
```
git clean -xfd
python setup.py sdist --formats=gztar
```

grlee77:
This resolves #249 (at least once we push a new source .tar.gz) to PyPI. I don't think we can remove the existing one, though? If not, we may have to tag a 0.11.1 release

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -558,7 +558,7 @@ def make_sniffer(compiler):
 
 def get_extensions():
     ext_modules = [Extension('pyfftw.pyfftw',
-                             sources=[os.path.join(os.getcwd(), 'pyfftw', 'pyfftw.pyx')])]
+                             sources=[os.path.join('pyfftw', 'pyfftw.pyx')])]
     return ext_modules
 
 


hgomersall:
I actually think there might have been a reason for the original behaviour, but i'm not sure...

hgomersall:
Ach, let's commit and see...

jakirkham:
Thanks for figuring this out and fixing it, @grlee77 😄

PR #251, opened 2018-11-19
remove pypi deployment in favor of pyFFTW-wheels

grlee77:
removes PyPI deployment from `appveyor/deploy.ps1`. This failed to work during the 0.11.0 release and we can soon upload wheels generated from pyFFTW-wheels instead (see pyFFTW/pyFFTW-wheels#24).


grlee77:
I cancelled the Travis build here as this only applies to Appveyor.

Issue #220, opened 2018-03-16
Add conda-forge badges


jakirkham:
There are badges for conda-forge, which would be nice to include in the README.

Should help users looking to install a pyFFTW binary find them quicker. Mentioning as there was no smaller number of issues about how to build/install pyFFTW. So providing a quick link to one easy way to get started would help. Adding wheels is another.

[Shields.io]( http://shields.io ) nicely provides them and one of our conda-forge contributors is to thank for the addition to their service. Here's what [they look like]( https://github.com/conda-forge/pyfftw-feedstock/blob/512782365bfcefe4f2322d5959d03ea182f43ef7/README.md#current-release-info ). Those are in Markdown, but we can use reStructuredText instead. Would be good to include them in our README.

Am a little unsure what sort of layout we would like to do. Feedback welcome. :)

hgomersall:
Yeah, let's have all the badges. Layout is not particularly important - as long as it's clear!

There is no reason the README uses rst other than for historical reasons. Now I'd use markdown, so if you want to change to MD, that's fine too!

PR #257, opened 2018-12-08
Update install instructions in README

grlee77:
This PR updates some of the install/build instructions in the README.

I did notice that it mentioned 32-bit builds, but we did not build these for the 0.11.1 release. 

I also added the additional conda-forge badges at the top as suggested in #220. ~I chose to keep the README as RST as opposed to switching to Markdown as I don't know that you can make a table like the one at the top in Markdown?~


grlee77:
Click on "View File" under the "Files chaned" section of this PR to see the rendered version.

PR #259, opened 2019-03-22
Avoid using "_N".

DWesl:
It is a preprocessor token on my platform, so builds fail with
unhelpful error messages.

### DIFF
--- a/pyfftw/pyfftw.pyx
+++ b/pyfftw/pyfftw.pyx
@@ -799,7 +799,7 @@ cdef class FFTW:
     cdef int64_t *_axes
     cdef int64_t *_not_axes
 
-    cdef int64_t _N
+    cdef int64_t _total_size
 
     cdef bint _normalise_idft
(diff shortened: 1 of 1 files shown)


hgomersall:
What's the objection to N?

DWesl:
I have no problems with renaming _N to N if you would prefer that, but I wanted to keep the leading underscore so I didn't accidentally imply it was part of the public API.

In case I misunderstood your question, my objection to the variable name "_N" is that something `#define`s that token to an integer on Cygwin and possibly all Newlib platforms, so I get errors when gcc has its turn with the generated C code.

hgomersall:
Ah, that's a good reason.

hgomersall:
I think it's rerunning the tests because I accidentally closed them. I think they passed, but we'll wait before merging. @grlee77 any objections to this (IMO trivial) change?

grlee77:
Looks good to me as well. I will go ahead and merge

PR #265, opened 2019-06-20
remove import of non-public _fftpack within the scipy interfaces

grlee77:
This PR removes the import of the non-public `_fftpack` submodule.

SciPy is currently in the process of adding pypocketfft as a replacement for FFTPack in scipy/scipy/#10238. At one point during file reorganization this broke the PyFFTW scipy interfaces because the `_fftpack` module had been moved. 

We later decided to leave it in its original location, but in either case, I don't think pyFFTW should be importing this non-public part of the API.

@peterbell10:  I assume this is the source of the issue you mentioned to me or was there something else?

On a related note, that same SciPy PR adds a `scipy.fft` module that is going to be encouraged for new code (although the `scipy.fftpack` API will be preserved for backwards compatibility). We can discuss that along with the new `scipy.fft` backend design later in a separate issue once it is a bit farther along.


peterbell10:
Yes, this was the source of the issue.

grlee77:
I plan to merge this by Monday if there are no objections.

PR #267, opened 2019-07-12
DOC: scipy_fftpack does not treat dtypes differently from scipy.fftpack

peterbell10:
With the possible exception of `np.longfloat`, the behaviour described here is already the behaviour of `scipy.fftpack`. I suspect this was just a copy-paste error from the `numpy_fft` interface docs.

### DIFF
--- a/pyfftw/interfaces/scipy_fftpack.py
+++ b/pyfftw/interfaces/scipy_fftpack.py
@@ -42,14 +42,6 @@ The exceptions raised by each of these functions are mostly as per their
 equivalents in :mod:`scipy.fftpack`, though there are some corner cases in
 which this may not be true.
 
-It is notable that unlike :mod:`scipy.fftpack`, these functions will
-generally return an output array with the same precision as the input
-array, and the transform that is chosen is chosen based on the precision
-of the input array. That is, if the input array is 32-bit floating point,
-then the transform will be 32-bit floating point and so will the returned
-array. Half precision input will be converted to single precision.  Otherwise,
-if any type conversion is required, the default will be double precision.
-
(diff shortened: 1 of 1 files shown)


peterbell10:
Apparently I accidently deleted the branch? Restored now anyway.

grlee77:
hmm... why did this just get closed/deleted again?

peterbell10:
I accidentally set my fork's remote to mirror this repo so whenever I pushed to #269 it would delete this branch.

hgomersall:
What's wrong with Appveyor?

peterbell10:
Looks like some dependencies are missing
```python
PackagesNotFoundError: The following packages are not available from current channels:
  - scipy=0.16
  - cython=0.23
  - numpy=1.10
```

A version bump might be enough to fix it.

hgomersall:
It's pulling things from conda, so I guess it depends on what that offers...

grlee77:
Yeah. I wonder if those were coming from the old `free` conda channel as opposed to `defaults` ([see here](https://www.anaconda.com/why-we-removed-the-free-channel-in-conda-4-7/))? IIRC, I also had problems getting this working last time I did a release and apparently something has changed again. 

Rather than spending time fixing this Windows+Python 2.7 case, I thought it may be better to just drop 2.7 support going forawrd as that seems to be what many other projects are currently doing. What do you think?

grlee77:
See #271 for an attempt at bumping the package versions used by Appveyor

grlee77:
The Appveyor issue was resolved in #271, but I'm not going to rerun the CI here as this PR involves only minimal documentation changes.

PR #269, opened 2019-07-12
scipy.fft interface

### LINKED ISSUE #268: Add a scipy.fft interface
Upcoming SciPy 1.4 will have a new `scipy.fft` module that is based on [pypocketfft](https://gitlab.mpcdf.mpg.de/mtr/pypocketfft). While `scipy.fftpack` will stick around, new code is encouraged to use `scipy.fft` which has some nice features as described in the following release note:
https://github.com/scipy/scipy/wiki/Release-note-entries-for-SciPy-1.4.0#scipyfft-added

see PRs:
scipy/scipy#10238
scipy/scipy#10361

As part of this effort, SciPy will also be gaining an official mechanism to allow setting different backends for the FFT computations. This will allow third party libraries like pyFFTW to define their own `scipy.fft`-compatible backend, removing the need for hacks like monkey patching:
scipy/scipy#10383
Peter (@peterbell10) will likely make a PR to pyFFTW in the near future with a draft of a scipy_fft interface

peterbell10:
Closes #268

Since the `scipy.fft` interface matches `numpy.fft` very closely, this is simple enough to implement.  The exception is the n-dimensional `hfft`s which would require more work than the 1d `hfft` in the numpy interface

### DIFF
--- a/docs/source/tutorial.rst
+++ b/docs/source/tutorial.rst
@@ -124,7 +124,7 @@ backend to speed up :func:`scipy.signal.fftconvolve`.
 
    t = Timer(lambda: scipy.signal.fftconvolve(a, b))
 
-   print('Time with scipy.fft: %1.3f seconds' % t.timeit(number=100))
+   print('Time with scipy.fft default backend: %1.3f seconds' % t.timeit(number=100))
 
    # Configure PyFFTW to use all cores (the default is single-threaded)
    pyfftw.config.NUM_THREADS = multiprocessing.cpu_count()
@@ -136,15 +136,15 @@ backend to speed up :func:`scipy.signal.fftconvolve`.
         # Turn on the cache for optimum performance
         pyfftw.interfaces.cache.enable()
 
-        print('Time with monkey patched scipy_fftpack: %1.3f seconds' %
+        print('Time with pyfftw backend installed: %1.3f seconds' %
                t.timeit(number=100))
 
 which outputs something like:
 
 .. code-block:: none
 
-   Time with scipy.fft: 0.598 seconds
-   Time with monkey patched scipy_fft: 0.251 seconds
+   Time with scipy.fft default backend: 0.598 seconds
+   Time with pyfftw backend installed: 0.251 seconds
 
 Prior to SciPy 1.4 it was necessary to monkey patch the libraries
 directly. :mod:`pyfftw.interfaces.numpy_fft` and
--- a/docs/source/tutorial.rst
+++ b/docs/source/tutorial.rst
@@ -96,26 +96,23 @@ which can be set with :func:`pyfftw.interfaces.cache.set_keepalive_time`,
 then they are removed from the cache (liberating any associated memory).
 The default keepalive time is 0.1 seconds.
 
-Monkey patching 3rd party libraries
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Integration with 3rd party libraries
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Since :mod:`pyfftw.interfaces.numpy_fft` and
-:mod:`pyfftw.interfaces.scipy_fftpack` are drop-in replacements for their
-:mod:`numpy.fft` and :mod:`scipy.fftpack` libraries respectively, it is
-possible to use them as replacements at run-time through monkey patching. Note
-that the interfaces (and builders) all currently default to a single thread.
-The number of threads to use can be configured by assigning a positive integer
-to `pyfftw.config.NUM_THREADS` (see more details under
-:ref:configuration <interfaces_tutorial>).
+SciPy versions 1.4 and above have support for installing different FFT
+backends. :mod:`pyfftw.interfaces.scipy_fft` support the use as a backend. Note
+that the interfaces (and builders) all currently default to a single thread. The
+number of threads to use can be configured by assigning a positive integer to
+`pyfftw.config.NUM_THREADS` (see more details under :ref:configuration
+<interfaces_tutorial>). The following code demonstrates using the :mod:`pyfftw`
+backend to speed up :func:`scipy.signal.fftconvolve`.
 
-The following code demonstrates :func:`scipy.signal.fftconvolve` being monkey
-patched in order to speed it up.
-
-.. testcode::
+.. codeblock:: python
 
    import pyfftw
    import multiprocessing
    import scipy.signal
+   import scipy.fft
    import numpy
    from timeit import Timer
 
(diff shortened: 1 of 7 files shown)


grlee77:
The CI failure here is unrelated (it seems availability of some conda packages for Python 2.7 have changed).

It may be time to just drop Python 2.7 support on master rather than try to fix this, but I will open a separate issue for that.

pyfftw/interfaces/__init__.py
```diff
@@ -116,6 +116,23 @@
...
+* :func:`pyfftw.interfaces.scipy_fft.ifft2`
+* :func:`pyfftw.interfaces.scipy_fft.fftn`
+* :func:`pyfftw.interfaces.scipy_fft.ifftn`
+* :func:`pyfftw.interfaces.scipy_fft.rfft`
+* :func:`pyfftw.interfaces.scipy_fft.irfft`
+* :func:`pyfftw.interfaces.scipy_fft.rfft2`
+* :func:`pyfftw.interfaces.scipy_fft.irfft2`
+* :func:`pyfftw.interfaces.scipy_fft.rfftn`
+* :func:`pyfftw.interfaces.scipy_fft.irfftn`
+* :func:`pyfftw.interfaces.scipy_fft.next_fast_len`
```
grlee77: It looks like `hfft` and `ihfft` are missing from this list

peterbell10: Thanks, fixed

grlee77:
Once uarray backend support has been added here, please also revise the following section of the docs:
 
https://github.com/pyFFTW/pyFFTW/blob/master/docs/source/tutorial.rst#monkey-patching-3rd-party-libraries

peterbell10:
`uarray` support is added and works for me locally. I've updated the tutorial as well.

hgomersall:
What're the uarray changes? I'm a bit out of touch on this...

peterbell10:
It was just merged into scipy today (scipy/scipy#10383). It means you can provide the implementation for the functions in `scipy.fft` without monkey-patching. e.g.
```python
import scipy.fft
scipy.fft.fft([1])  # Calls scipy's own implementation

from pyfftw.interfaces import scipy_fft
scipy.fft.set_global_backend(scipy_fft)
scipy.fft.fft([1])  # Calls into pyfftw
```
See the [`uarray` docs] for a description of `__ua_function__`. It's very similar to NumPy's new `__array_function__` protocol if you've heard of that.

[`uarray` docs]: https://uarray.readthedocs.io/en/latest/libauthor_docs.html

hgomersall:
Does this break without the latest scipy? Is it only this set backend bit?

peterbell10:
The only issue is that the symbol `scipy.fft.set_backend` doesn't exist and it's used in the test. I could just skip that test and everything should be fine.

peterbell10:
And of course if `scipy.fft` doesn't exist then everything is excluded and skipped already.

hgomersall:
I think I'd prefer the import code to inspect the scipy version to make it explicit for future reading (`scipy.__version__`). You could use the same with a SkipIf decorator on the test that's failing, then it should work. Do the dev versions bump the version at the beginning of a development cycle?

peterbell10:
>I think I'd prefer the import code to inspect the scipy version to make it explicit for future reading (scipy.__version__).

Done

>You could use the same with a SkipIf decorator on the test that's failing, then it should work.

Unfortunately the weekly builds are only identified by part of a commit hash so ordered comparison isn't really possible.

>Do the dev versions bump the version at the beginning of a development cycle?

Yes, it's currently on 1.4.0.dev0 while the released version is 1.3.0

hgomersall:
> Unfortunately the weekly builds are only identified by part of a commit hash so ordered comparison isn't really possible.

Doesn't the version convey the information? I thought the api changes was 1.3 -> 1.4?

peterbell10:
Yes, for release versions that will be sufficient but 1.4 hasn't been released yet.

hgomersall:
Do the comparisons not work on dev releases?

peterbell10:
The dev versions are like `1.4.0.dev0+<commit hash>` and hashes aren't ordered so comparing hashes with `>=` doesn't make sense. You can tell that it is a `1.4.0` dev release but you can't tell how far into development it is.

hgomersall:
Ah I see. Anyway, it looks fine to me. The problem at the moment then is that is breaks against dev releases, but should work find on non-dev versions?

peterbell10:
Yes, it's only an issue with 1.4.0dev versions from before the uarray features were added. Notice that on [travis] everything else is passing.

[travis]: https://travis-ci.org/pyFFTW/pyFFTW/builds/560988974?utm_source=github_status&utm_medium=notification

hgomersall:
Ok, in that case let's wait for the weekly release and then merge.

hgomersall:
Nice work by the way :)

larsoner:
@peterbell10 if it makes life easier for this backend work we can probably increment the `dev` number as `1.4.0dev1`

peterbell10:
Unfortunately `LooseVersion` isn't smart enough to realise `1.4.0 > 1.4.0.dev0` which would complicate the checking further. I think just waiting until Monday is good enough.

peterbell10:
The new wheels are out and the tests are now passing.

pyfftw/interfaces/scipy_fft.py
```diff
@@ -0,0 +1,342 @@
...
+        planner_effort=None, threads=None,
+        auto_align_input=True, auto_contiguous=True):
+    '''Perform an n-D real FFT.
+
+    The first five arguments are as per :func:`scipy.fft.rfftn`;
+    the rest of the arguments are documented
+    in the :ref:`additional argument docs<interfaces_additional_args>`.
+    '''
+    x = np.asanyarray(x)
+    if not np.isrealobj(x):
```
grlee77: Using `if np.iscomplexobj(x)` should have slightly less overhead
 (`isrealobj` calls `iscomplexobj` internally)

Alternatively, `if x.dtype.kind == 'c':` should be even faster than calling `np.iscomplexobj`.

test/test_pyfftw_scipy_fft.py
```diff
@@ -0,0 +1,144 @@
+# Copyright 2014 Knowledge Economy Developments Ltd
+#
+# Henry Gomersall
+# heng@kedevelopments.co.uk
```
grlee77: same here about using "PyFFTW developers"

docs/source/tutorial.rst
```diff
@@ -127,35 +124,41 @@ patched in order to speed it up.
...
+               t.timeit(number=100))
 
 which outputs something like:
 
 .. code-block:: none
 
-   Time with scipy.fftpack: 0.598 seconds
-   Time with monkey patched scipy_fftpack: 0.251 seconds
+   Time with scipy.fft default backend: 0.598 seconds
+   Time with pyfftw backend installed: 0.251 seconds
```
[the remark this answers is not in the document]
peterbell10: The gap has closed slightly but I think the main advantage pyfftw has now is multithreading.

grlee77:
@hgomersall, If these changes look good to you I think we are done here.

PR #274, opened 2019-09-09
Update scipy.fft interface to fix shape handling and add workers argument

grlee77:
This PR updates the recently introduced `scipy.fft` interface so that it is in sync with SciPy master.

1.) This adds the `workers` argument to the `scipy.fft` interface to match scipy/scipy#10614 

2.) updates shape handling as in scipy/scipy#10594 (we can just fall back to the numpy_fft interface shape handling now)

Note: The `workers` argument here replaces the functionality previously provided by the `threads` argument. As we have never tagged a release with the new `scipy.fft` interface yet, I don't think this is a problem for backwards compatibility. 

The Travis case using the SciPy development wheels will be the relevant one to monitor here as SciPy 1.4 (which introduces the `scipy.fft` module) has not yet been released.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -125,7 +125,7 @@ html_theme = 'default'
 # Add any paths that contain custom static files (such as style sheets) here,
 # relative to this directory. They are copied after the builtin static files,
 # so a file named "default.css" will overwrite the builtin "default.css".
-html_static_path = ['_static']
+html_static_path = []
 
 # If not '', a 'Last updated on:' timestamp is inserted at every page bottom,
 # using the given strftime format.
--- a/docs/source/tutorial.rst
+++ b/docs/source/tutorial.rst
@@ -510,7 +510,7 @@ and planning effort.
 
 The initial values in pyfftw.config at import time can be controlled via the
 environment variables as detailed in the
-:ref:`configuration <_configuration_variables>` documentation.
+:ref:`configuration <configuration_variables>` documentation.
 
 .. rubric:: Footnotes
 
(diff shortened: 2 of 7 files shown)


peterbell10:
I think the "Additional parameters" section of the docs needs to be updated as well:
https://github.com/pyFFTW/pyFFTW/blob/85ff8bd22a29a07a6ee935847a601d8efa8251d7/pyfftw/interfaces/__init__.py#L216-L218

grlee77:
> I think the "Additional parameters" section of the docs needs to be updated as well:

Right, good catch. 
The `threads` parameter is still an "Additional parameter" for the numpy and scipy.fftpack interfaces so I don't think we want to remove it from this section. Perhaps we should leave it here, but put a note that this "additional" parameter is superseded by the `workers` argument in the case of the `scipy.fft` interfaces?

peterbell10:
> Perhaps we should leave it here, but put a note that this "additional" parameter is superseded by the `workers` argument in the case of the `scipy.fft` interfaces?

Yes, that's what I was meaning. It should be updated to match the note currently found on `overwrite_input`:

https://github.com/pyFFTW/pyFFTW/blob/85ff8bd22a29a07a6ee935847a601d8efa8251d7/pyfftw/interfaces/__init__.py#L188-L190

grlee77:
@peterbell10: see how the text in aa5edc6 looks to you. (The other new  commits just fix a few Sphinx warnings I encountered, but are not specific to the changes introduced in this PR)

grlee77:
Hi @hgomersall, can you look at potentially merging this soon?

It is an update to #269 that makes the interface consistent with what we ended up with in SciPy, including a `workers` argument for specifying the number of threads.

hgomersall:
I noticed the removal of the check on the shape. I assume this has been sorted in SciPy?

grlee77:
Yes, the shape handling was made consistent with NumPy in scipy/scipy#10594 so we no longer need the spatial case here.

PR #277, opened 2019-11-21
Try to bump against dask 2.8.0

### LINKED ISSUE #276: futurewarnings in dask 2.8 break test suite
please add dask 2.8 compat

seems like only the test suite is affected.

I had to ignore futurewarnings coming from dask, but I guess that's only a temporary solution.

xnox:
Trying to fix #276

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,8 @@ matrix:
   include:
     - python: 3.7
       env: SCIPY_WHEELS="https://7933911d6844c6c53a7d-47bd50c35cd79bd838daf386af554a83.ssl.cf2.rackcdn.com/"
+    - python: 3.8
+      env: SCIPY_WHEELS="https://7933911d6844c6c53a7d-47bd50c35cd79bd838daf386af554a83.ssl.cf2.rackcdn.com/"
 
 branches:
     except:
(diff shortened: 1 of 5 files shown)


hgomersall:
Do you know why #276 is happening?

xnox:
@hgomersall 
It seems to be caused by dask 2.8.0, see the CI results on this PR that also trigger it (modified to try to use dask 2.8.0 here). I.e. https://travis-ci.org/pyFFTW/pyFFTW/builds/615315904?utm_source=github_status&utm_medium=notification

It does introduce new FutureWarnings that do trip up a few test cases that use `numpy.alltrue` as that is not implemented by dask.
I don't know why `test_shape_and_s_different_lengths` with dask fails.

Also it looks like it is passing with 3.8, but fails under 3.7?

And I failed to make AppVeyor work with 3.8.

xnox:
In Ubuntu, the same test case does fail with python3.8 as seen in https://launchpadlibrarian.net/452580238/buildlog_ubuntu-focal-amd64.pyfftw_0.11.1-3ubuntu1_BUILDING.txt.gz

grlee77:
closing as this PR was superseded by #278

PR #278, opened 2019-11-25
Update Dask tests for compatibility with recent Dask

### LINKED ISSUE #276: futurewarnings in dask 2.8 break test suite

grlee77:
closes #276

Trying this first without bumping dask to 2.x to see if it maintains compatibility with old versions.
Then I can try the dask 2.8 bump as suggested in #277

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,22 +1,30 @@
 language: python
 
-sudo: false
-
 env:
     global:
         # github_changelog_generator API key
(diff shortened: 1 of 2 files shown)


grlee77:
ping @jakirkham as the original author of these tests. Do these changes look reasonable to you? (I am not a frequent dask user)

jakirkham:
Yeah this seems fine. Thanks for updating them @grlee77! 😀

grlee77:
Do you want to keep the Travis changes in the same PR. I dropped Python 3.4 to save time because it has reached "end of life" and we probably don't want to keep supporting it going forward.

I should restore SCIPY_WHEELS in one case here before merging so that we still test against upcoming scipy 1.4 for the new `scipy.fft` interfaces.

jakirkham:
Fine by me. Maybe @hgomersall has thoughts?

grlee77:
Actually, looking more closely at the logs, dask is only being tested for Python 2.7 and 3.5. I will have to update the version to a more recent one for a conda package to exist for the newer Python versions

hgomersall:
@jakirkham This all looks fine to me. I don't use Dask, so I'll have to leave the usage issues to you. I'm probably not the best person to review, but I'm happy for it to be merged once people are happy.

grlee77:
@hgomersall or @jakirkham: can one of you merge this, please? It should be good to go.

hgomersall:
@grlee77 You can merge this now...

PR #283, opened 2019-12-27
explicitly set the Cython language level

grlee77:
This avoids a warning when building with recent Python.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,7 +45,7 @@ before_install:
   - conda info -a
 
 install:
-  - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION nomkl numpy cython setuptools
+  - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION nomkl numpy setuptools
   - conda install -q -n test-environment $DASK_SPEC || true
   - |
     if [ -z "$SCIPY_WHEELS" ]; then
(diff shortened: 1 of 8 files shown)


grlee77:
It looks like 3str was only introduced in Cython 0.29, so we should either set it to 2 or update the minimum Cython version required.

grlee77:
Okay, the CI was a little bit of a pain here, but hopefully it is going to pass this time around.

The issue addressed in this PR is that whenever Cython 3.0 is released it will default to language level `3str` which is basically Python 3 semantics, but with Python 2 `str` support. As pyFFTW was using the older `language_level=2` semantics, this would break pyFFTW under Cython 3. Given that, I think we definitely should specify the language level. There are 3 ways of going about this

1. Keep the current Cython code and specify `language_level=2` everywhere
2. update Cython code to use Python 3 semantics, but set `langauge_level=3str` to retain Python 2 compatibility.
3. update Cython code to use Python 3 semantics, and set `language_level=3`, dropping Python 2.7 support

I did approach 2 in this PR. Switching to approach 3 should be trivial once we drop Python 2.7.

I propose we drop 2.7 and 3.5 support immediately after the new release and switch to `language_level=3` at that time. What do you think?

grlee77:
I'm not sure why the extra commit e247383 related to the dask tests was in this PR. I can rebase and remove it. Let's see if the CI passes first, though.

hgomersall:
Oh, the changes were tiny - good news! I think we can drop 2.7 support now everybody else is. It's time to move on. In which case you proposal sounds good.

grlee77:
@hgomersall, If you approve this one, I can merge once the CI is green.

After that I think we should be good to tag the 0.12 release and then create a PR over at pyFFTW-wheels to finish building all wheels for the release.

PR #285, opened 2020-02-03
Drop testing on Python 2.7 and 3.5

### LINKED ISSUE #270: Plan for dropping Python 2.7 support
We should specify a plan for dropping Python 2.7 support. 

Numpy [dropped support](https://docs.scipy.org/doc/numpy-1.14.2/neps/dropping-python2.7-proposal.html) for 2.7 (and 3.4) in new releases starting at the beginning of 2019 and have only committed to making bug fixes to the 1.16.x branch supporting Python 2.7 until Jan 1, 2020. Many other packages in the ecosystem have dropped 2.7 support at this point as well.

I would be in favor of a similar approach here, dropping Python 2.7 support from master going forward. If we need/want to make a bug fix release for 0.11.x that maintains 2.7 support we can always create a branch for that.

Dropping 2.7 from our CI would also avoid needing to fix some recent failures on Windows due to lack of conda packages as seen in #269.

grlee77:
closes #270

Python 3.5 should continue to work for now, but I don't think we need to continue officially supporting it.

Many related packages may now be following NumPy's [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) guidance and will have already dropped Python 3.5 support at this point.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ The documentation can be found on
 
 ## Requirements (i.e. what it was designed for)
 
-- [Python](https://python.org) 2.7 or >= 3.4
+- [Python](https://python.org) >= 3.6
 - [Numpy](https://www.numpy.org) >= 1.10.4  (lower versions *may* work)
 - [FFTW](https://www.fftw.org) >= 3.3 (lower versions *may* work) libraries for
   single, double, and long double precision in serial and multithreading
--- a/README.md
+++ b/README.md
@@ -45,8 +45,8 @@ The documentation can be found on
 
 ## Requirements (i.e. what it was designed for)
 
-- [Python](https://python.org) >= 3.6
-- [Numpy](https://www.numpy.org) >= 1.10.4  (lower versions *may* work)
+- [Python](https://python.org) >= 3.7 (lower versions *may* work)
+- [Numpy](https://www.numpy.org) >= 1.16 (lower versions *may* work)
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -45,7 +45,7 @@ The documentation can be found on
 
 ## Requirements (i.e. what it was designed for)
 
-- [Python](https://python.org) 2.7 or >= 3.4
+- [Python](https://python.org) >= 3.6
```
hgomersall, line 48: It would be good to add a comment that previous versions supported 2.7 and it *may* still work.

grlee77: looking again, I see that we already have a statement to this effect a few lines further down:
```
In practice, pyFFTW *may* work with older versions of these dependencies, but it is not tested against them.
```

grlee77: I can add it in line as well if you like. I see that NumPy has that stated in-line

grlee77:
Also, if we are going to continue using Travis, you will need to migrate the repository from travis-ci.org (which is shutting down) to travis-ci.com as described here: https://docs.travis-ci.com/user/migrate/open-source-repository-migration

I  do not have the permissions to change settings for PyFFTW on Travis

hgomersall:
Looks good. Once the tests pass we'll merge. I think I'll worry about the travis-ci.org change later on - I'm super hectic at the moment! Moving to github might be a good idea.

hgomersall (approved):

hgomersall:
Thanks @grlee77!

PR #305, opened 2021-02-05
Minor doc changes to README

hgomersall:
@grlee77 would you mind sanity checking these minor doc changes?

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,7 +160,7 @@ suitable ``.lib`` files as described on the
 
 #### Mac OSX
 
-Install FFTW from [homebrew](http://brew.sh>)::
+Install FFTW from [homebrew](http://brew.sh):
 
   brew install fftw
 
--- a/README.md
+++ b/README.md
@@ -73,8 +73,15 @@ package index with tools such as pip:
 
     pip install pyfftw
 
-Pre-built binary wheels for 64-bit Linux, Mac OS X and Windows are available on
-the [PyPI](https://pypi.org/) page for all supported Python versions.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -73,8 +73,15 @@ package index with tools such as pip:
...
-the [PyPI](https://pypi.org/) page for all supported Python versions.
+Pre-built binary wheels for 64-bit Python on  Linux, Mac OS X and Windows are 
+available on the [PyPI](https://pypi.org/) page for all supported Python versions.
+Note that we only support binaries for 64-bit python. If you need to use 32-bit
+python for some reason you will have to build youtself.
+
+Note that prior to Python 3.9, the windows installation defaulted to being 32-bit 
+even on 64-bit Windows, so if you are having problems installing using pip
+(typically with an error message like `ERROR: Failed building wheel for pyfftw`)
+then please check your python version.
```
grlee77, line 84:
```suggestion
Pre-built binary wheels for 64-bit Python on Linux, Mac OS X and Windows are 
available on the [PyPI](https://pypi.org/) page for all supported Python versions.
Note that we only support binaries for 64-bit Python. If you need to use 32-bit
python for some reason, you will have to build pyFFTW from source.

Note that prior to Python 3.9, the Windows installation defaulted to being 32-bit 
even on 64-bit Windows, so if you are having problems installing using pip
(typically with an error message like `ERROR: Failed building wheel for pyfftw`)
then please check your Python version.
```

grlee77 (approved):
It looks good to me aside from some minor edits. I will mark as approved now so you can go ahead and merge once you check those

PR #256, opened 2018-12-01
Real-to-real transforms (version 3)

grlee77:
I have rebased the nice work by @drwells from #95 on current master, so the first 20 or so commits here are the ones from that PR with modifications made only as necessary to resolve conflicts.

That initial rebase went fairly quickly, but then it took an hour or two to track down a few mistakes I made  (fixed in 527c4e3 and d7ee369).

I updated the `threads` and `planner_effort` handling to be consistent with pyFFTW 0.11+ and added the n-dimensional variants of the DCT and DST introduced in SciPy 1.0. Currently these are implemented by separable application of 1D transforms rather than trying to implement an n-dimensional R2R builder.

I have copied over the TODO list from a comment in #95 below for quick reference:

- [ ] Tests for builder functions
- [ ] Tests for failure modes
- [x] Improve SciPy interface's handling of `None` with orthogonalization
- [ ] Implement padding and truncation for the `dct` and `dst` functions
- [x] Clean up scipy tests
- [x] Clean up real to real tests (use existing infrastructure)
- [ ] Improve `_Xfftn`
- [x] Update documentation to describe real to real features 

New items I added
- [x] flexible defaults for threads and planner_effort in the builders and interfaces
- [x] dctn, idctn, dstn, dctn added to interfaces.scipy_fftpack
- [x] test axis argument of dct, etc. in the scipy interfaces tests
- [x] add tests for n-dimensional R2R transforms in the scipy interfaces
- [ ] n-dimensional builders for dctn, dstn?

If anyone want to volunteer to work on any of the remaining items, please do so.

PR #308, opened 2021-03-13
ENH: update interfaces to support new numpy.fft normalization options

### LINKED ISSUE #295: update interfaces to support new normalization options
This PR is a reminder that we will likely need to update the numpy and scipy interfaces to support additional `norm` options after a future NumPy release.

Currently the only options are:
1.) 1/N normalization on the inverse transform  (when `norm=None`)
2.) 1/sqrt(N) on both (when `norm='ortho'`)

https://github.com/numpy/numpy/pull/16476 adds two additional keyword options.:
1.) norm='backward' is an alias for the current `norm=None` behavior and applies 1/N normalization during inverse transforms only
2.) norm='forward' applies 1/N normalization during the forward transforms only
see also: https://github.com/numpy/numpy/issues/16126

chrisvalx:
**ENH: update interfaces to support new numpy.fft normalization options**

Closes #295.

This PR extends the normalization options of the `pyfftw` builders and `numpy_interface`, so that they're up to date with the current `numpy.fft` options (see [numpy/numpy#16476](https://github.com/numpy/numpy/pull/16476)). In particular, the `norm` keyword argument now accepts the following values:
- `norm=backward` (alias of `norm=None`, default case) which scales the backward FFT with 1/N and leaves the forward unscaled;
- `norm=ortho` which scales both directions with 1/sqrt(N);
- `norm=forward` which scales the forward FFT with 1/N and leaves the backward unscaled. 

pyFFTW uses two boolean variables `ortho` and `normalise_idft` to determine the correct normalization, instead of one string variable `norm` like in NumPy. The main change is in the builders utilities function `_norm_args(norm)` which receives `norm` and outputs a dictionary with the correct scaling flags. Both the builders and the interfaces use this function. `_Xfftn` has also been modified to apply the correct normalization after the FFT has been executed. The `dct` and `dst` builders don't accept any normalization options other than the default, so no changes made there.

The main change in the NumPy interface is in the hermitian routines `hfft` and `ihfft`; due to symmetry, `hfft` is done via `irfft` and `ihfft` via `rfft`, which means that before calling the real FFTs we need to swap direction of the normalization flags. For that I've added the `_swap_direction_dict` module variable and the `_swap_direction(norm)` function, which make sure the "real" norm value is established correctly; that's now handled in the exact same way as in NumPy.

I've added tests in the builders testsuite that check the new norm options are called correctly, and that roundtrip equality holds for all norm values (i.e. `ifft(fft(a)) == a` up to numerical accuracy for all `norm` values). I've added those in the FFTW wrapper tests, since they require both forward and backward FFTs. For the NumPy interface I've added test cases for the new `norm` options, but couldn't figure out how to check for roundtrip consistency. The interfaces use the same underlying routines already checked by the builders tests, but it'd be nice to explicitly check the interface calls as well.

All tests pass at the time of writing, but I'm not sure my approach is correct so if someone more experienced could take a look at the tests that'd be appreciated. Just so that we're sure all cases are covered properly.

### DIFF
--- a/pyfftw/interfaces/cache.py
+++ b/pyfftw/interfaces/cache.py
@@ -70,14 +70,14 @@ is not available and trying to use it will raise an ImportError exception.
 
 The actual implementation of the cache is liable to change, but the
 documented API is stable.
-'''
 
+'''
 try:
     import threading as _threading
     _threading_import_error = None
 except ImportError as e:
     _threading_import_error = e
-    _threading == None
+    _threading = None
 
 import time
 import weakref
@@ -86,8 +86,10 @@ __all__ = ['enable', 'disable', 'set_keepalive_time']
 
 _fftw_cache = None
 
+
 class CacheError(Exception):
     pass
+
 
 def enable():
     '''Enable the cache.
@@ -100,11 +102,13 @@ def enable():
     else:
         raise ImportError(_threading_import_error)
 
+
 def disable():
     '''Disable the cache.
     '''
     global _fftw_cache
     _fftw_cache = None
+
 
 def is_enabled():
     '''Return whether the cache is currently enabled.
@@ -114,9 +118,11 @@ def is_enabled():
     else:
         return True
 
+
 def set_keepalive_time(keepalive_time):
-    '''Set the minimum time in seconds for which any :mod:`pyfftw.FFTW` object
-    in the cache is kept alive.
+    '''
+    Set the minimum time in seconds for which any :mod:`pyfftw.FFTW`
+    object in the cache is kept alive.
 
     When the cache is enabled, the interim objects that are used through
     a :mod:`pyfftw.interfaces` function are cached for the time set through
@@ -125,8 +131,9 @@ def set_keepalive_time(keepalive_time):
 
     The time is not precise, and sets a minimum time to be alive. In
     practice, it may be quite a bit longer before the object is
-    deleted from the cache (due to implementational details - e.g. contention
-    from other threads).
+    deleted from the cache (due to implementational details - e.g.
+    contention from other threads).
+
     '''
     global _fftw_cache
 
@@ -134,6 +141,7 @@ def set_keepalive_time(keepalive_time):
         raise CacheError('Cache is not currently enabled')
     else:
         _fftw_cache.set_keepalive_time(keepalive_time)
+
 
 class _Cache(object):
 
@@ -158,10 +166,11 @@ class _Cache(object):
         self._close_thread_now = _threading.Event()
 
         self._initialised = _threading.Event()
-        self._initialised.clear() # Explicitly clear it for clarity
+        self._initialised.clear()  # explicitly clear it for clarity
 
         self._thread_object = _threading.Thread(target=_Cache._run,
-                args=(weakref.proxy(self), ), name='PyFFTWCacheThread')
+                                                args=(weakref.proxy(self), ),
+                                                name='PyFFTWCacheThread')
 
         self._thread_object.daemon = True
         self._thread_object.start()
@@ -195,7 +204,7 @@ class _Cache(object):
 
             while True:
                 if (not self._parent_thread.is_alive() or
-                    self._close_thread_now.is_set()):
+                        self._close_thread_now.is_set()):
                     break
 
                 if time.time() - last_cull_time > self._keepalive_time:
@@ -227,12 +236,14 @@ class _Cache(object):
             pass
 
     def set_keepalive_time(self, keepalive_time=0.1):
-        '''Set the minimum time in seconds for which any object in the cache
+        '''
+        Set the minimum time in seconds for which any object in the cache
         is kept alive.
 
         The time is not precise, and sets a minimum time to be alive. In
         practice, it may be up to twice as long before the object is
         deleted from the cache (due to implementational details).
+
         '''
         self._keepalive_time = float(keepalive_time)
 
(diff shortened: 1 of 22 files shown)


pyfftw/interfaces/cache.py
```diff
@@ -70,14 +70,14 @@
...
-'''
 
+'''
 try:
     import threading as _threading
     _threading_import_error = None
 except ImportError as e:
     _threading_import_error = e
-    _threading == None
+    _threading = None
```
chrisvalx, line 80: should be `=` instead of `==` right?

chrisvalx:
The only test currently failing in my build is `test_objects_removed_after_keepalive` in file `test_pyfftw_interfaces_cache.py`, with error message from line 337:

```
AssertionError: KeyError not raised by lookup
```

pyfftw/interfaces/scipy_fftpack.py
```diff
@@ -412,9 +437,41 @@ def irfft(x, n=None, axis=-1, overwrite_x=False,
                            auto_contiguous)
 
 
+# why map None:None, and not None:"forward"?
+# if I change it to map None:"forward" some `scipy_interface` tests fail.
+# see also the if-clauses in the sine/cosine transforms in `scipy_fft.py`;
+# there's some mishandling of norm=None somewhere.
+_swap_norm_dictionary = {"backward": "forward", None: None,
```
chrisvalx, line 444: that mapping of `None:None` was like that in the code, and I'm not sure why it shouldn't map `None:"forward"`; I've left it as is for now because some of the tests were failing otherwise (see code comments)

pyfftw/interfaces/scipy_fft.py
```diff
@@ -370,8 +368,10 @@ def dct(x, type=2, n=None, axis=-1, norm=None, overwrite_x=False,
 
     """
     threads = _workers_to_threads(workers)
+    # not sure why's that if-clause necessary, but some `scipy_fft`
+    # tests fail otherwise
     if norm is None:
-        norm = 'backward'
+        norm = "backward"
```
chrisvalx, line 374: also not sure why that change is necessary in the `scipy_fft` sine/cosine transforms

hgomersall: This seems suspicious to me...

pyfftw/interfaces/cache.py
hgomersall: Yes, good catch. I guess the non-threading path never gets tested!

pyfftw/interfaces/scipy_fft.py
chrisvalx: both this if-clause and the mapping `None:None` referenced in my comment above seem to be related to the work done on #256.

I don't want to start changing too many things unrelated to the current PR, so I left them unchanged to make sure no tests are failing. Could be looked into with a new PR that focuses on changes leftover from #256 and similar work. 

@grlee77

hgomersall: Ok, but changing single quotes to double quotes shouldn't change anything.

chrisvalx: no the single-to-double quotes change was just done while I was doing aesthetic changes in bulk (unrelated); the change I'm referring to is why should one need to change `norm=None` to `norm="backward"` for sine/cosine transforms, even though they're defined as equivalent parameters.

Such a change is not necessary for any of the other FFTs.

chrisvalx:
some CI tests related to the `dask` interface seem to fail, because of warnings about using `numpy.alltrue()` among others; these tests pass just fine in my local build, only triggering non-failure warnings about this issue. 

This seems to have to do with the inners of `dask` and not the `pyfftw` wrappers around it, although I have no experience working with `dask`.

pyfftw/interfaces/scipy_fft.py
hgomersall: Yeah, not sure about that.

As an aside, it would be appreciated if you didn't make bulk minor stylistic changes - it makes parsing the PR much harder and is arguably undesirable. Changes to be PEP8 compliant are valued, but just changing single quotes to double quotes seems like unnecessary noise to me.

chrisvalx: you're right I'm sorry about that; I made them in bulk through my editor so I underestimated how noisy it makes the PR changes look

chrisvalx:
none of the CI build checks seem to run the tests with `numpy 1.20.0` or newer, which is the version that implemented the extended `norm` options; that's why build checks before commit b1be82d90637b517a4372a6fac8f66d9e964f393 were failing. In my local build I'm running tests with `numpy 1.20.1` and all tests pass as well.

hgomersall:
Great. Is this ready for review now?

chrisvalx:
Yes I believe so, the main work is done.

pyfftw/interfaces/scipy_fft.py
leofang: There's too much noise that is impossible for non-maintainers like me to help review (not saying I have time right now, but still)...(btw I made the counterpart change in CuPy: cupy/cupy#4797, cupy/cupy#4816) It'd be really nice if this PR can be split into two.

grlee77:
It would be good to get this in. Let me help rebase this after the tests folder was renamed.

grlee77:
Thanks @chrisvalx, I have updated this in #330 so we can close this one and proceed from there (you are still credited as a commit author).

Issue #317, opened 2021-09-30
Real-to-real transformations seem to be excluded from the latest release


lukastk:
In the API documentation for [pyfftw.builders](https://pyfftw.readthedocs.io/en/latest/source/pyfftw/builders/builders.html) version 0.12.0 the functions

- `pyfftw.builders.dct`
- `pyfftw.builders.dst`

are listed, but do not seem to be included in the latest pip release of pyfftw. Inspecting `pyfftw/builders/builders.py`in site-packages, these functions do seem to be missing (although they are present in the repository).

If I run the following snippet,

```python
d = pyfftw.empty_aligned(128, dtype='float64')
e = pyfftw.empty_aligned(128, dtype='float64')

dct_transform = pyfftw.FFTW(d, e, direction='FFTW_REDFT00')
```

taken from the [tutorial](https://pyfftw.readthedocs.io/en/latest/source/tutorial.html#one-dimensional-transforms) I get the error _Invalid direction: The direction is not valid for the scheme. Try setting it explicitly if it is not already_.

It seems that all the cosine and sine transformations are excluded from the latest public version of pyfftw. Apologies if I'm pointing out the obvious here. If nothing else I just wanted to point out the discrepancy between the documentation and the released version.

hgomersall:
Yeah, the release is old and needs to be updated soon...

grlee77:
A new release (0.13.0) has been published that has these functions

Issue #312, opened 2021-05-11
Tag v0.12.0 is not an ancestor of master


jwoillez:
Just a note to let you know that the current master branch does not have the v0.12.0 tag as an ancestor. `git describe` still gives `v0.11.1-66-gd74d032` rather than something derived from the current released `v0.12.0` version. This keeps the version of pyFFTW installed from master as based on `v0.11.1`.

grlee77:
Yeah, I must have messed this up when doing the 0.12.0 release. Not sure what we can do about it now, though? If we do a new release candidate soon, we need to make sure to use 0.13.0rc0!

grlee77:
The v0.13.0 tag should be an ancestor of master, so versions on new commits should now show up as something like `0.13.0+1.ge6e475c` (i.e. no longer erroneously starting from 0.11)

Issue #309, opened 2021-03-14
Add instructions for installing inside GitHub actions


cubranic:
I have a package that has pyFFTW as a dependency and I don't have any problems installing it except in Github actions. There, it fails with the same "unknown file type '.pyx'" error from #252 . The solution was to have the following lines in the installation workflow:
```
python -m pip install --upgrade pip wheel setuptools cython
sudo apt install -y libfftw3-dev
```
Is this the right way to go about it? Either way, it took a long time to figure it out by trial and error (or rather, commit-push-see output listing), and it would have been really useful to have that mentioned in the Readme.

hgomersall:
That looks reasonable to me. I guess the github actions couldn't find a pre-compiled wheel? Certainly, if you build from scratch then you'll need to have `libfftw3-dev` installed. For building, `FFTW` _is_ listed as a requirement. It's not clear to me why github actions needed to try to build from scratch. What is it looking for at PyPI?

cubranic:
Without installing anything extra and just having `pip install pyfftw`, I see the following in the action’s output:
```
Collecting pyfftw~=0.12.0
52
 <https://github.com/chime-sps/ps-processes/runs/2104580876#step:4:52>  Downloading pyFFTW-0.12.0.tar.gz (137 kB)
53
 <https://github.com/chime-sps/ps-processes/runs/2104580876#step:4:53>
Using legacy 'setup.py install' for pyfftw, since package 'wheel' is not installed.
73
 <https://github.com/chime-sps/ps-processes/runs/2104580876#step:4:73>
```
Even after I installed `wheel` first, it still tried to build and failed with no fftw3.h header. Then I added that `apt install libfftw3-dev` and got the `pyx` error. Only after I also added cython did it start working, and now I get:
```
  Building wheel for pyfftw (setup.py): started
219
 <https://github.com/chime-sps/ps-processes/runs/2107033727#step:4:219>  Building wheel for pyfftw (setup.py): finished with status 'done'
220
 <https://github.com/chime-sps/ps-processes/runs/2107033727#step:4:220>  Created wheel for pyfftw: filename=pyFFTW-0.12.0-cp39-cp39-linux_x86_64.whl size=530907 sha256=2c6b7fb685e296663a17563045abfe3c36a458d23fe4575de0735ecbefe73715
221
[...]

Issue #301, opened 2021-01-28
Python 3.9 support


hgomersall:
It's not currently implemented.

hgomersall:
How are Mac binaries defined? Also, anyone know about the M1?

HaHeho:
(I believe) I am using `fftw 3.3.9` from brew with `pyfftw 0.12.0` from conda-forge successfully in `Python 3.9.1` .

Not implemented in what sense?

hgomersall:
In the sense it isn't in PyPI or in any of the test matrices.

akx:
Is https://github.com/pyFFTW/pyFFTW-wheels the project currently used to build wheels for PyPI too? 

(I'm trying to build pyFFTW on macOS on Python 3.9 against a Homebrew FFTW to not much success ("Could not find any of the FFTW libraries"), so a wheel would be nice...)

hgomersall:
@grlee77 python 3.8 uses a scipy wheel build on a rackspace server. I vaguely remember something about this, but not the details. I think you implemented this bit...?

Just implementing the test matrix changes now...

grlee77:
I will have to go back and see what was complete here. We can potentially have multibuild upload the wheels to https://anaconda.org/multibuild-wheels-staging/, which is what I had done recently a couple of other places.

2-3 months ago Travis started limiting jobs to 10,000 free credits per month for open source projects. On linux, 1 minute = 10 credits, but I think OS X was 5 times more expensive. Given that we only have linux jobs on Travis and are not super active with new PRs, maybe 1000 minutes / month will be okay. If so we can probably keep things mostly as is just changing from travis-ci.org to travis-ci.com. For larger projects like SciPy and scikit-image these free credits get exhausted within a day or two, so they had to move away from using Travis.

Another alternative to [multibuild](https://github.com/matthew-brett/multibuild) is to use [cibuildwheel](https://cibuildwheel.readthedocs.io/en/stable/) with GitHub Actions and the wheels will show up directly in the Actions tab in this repository. I have not done this myself, but matplotlib and scikit-image are now using this approach: https://github.com/scikit-image/scikit-image/pull/5080

It is also possible to add aarch64 support via cibuildwhel on linux: https://github.com/scikit-image/scikit-image/pull/5197

grlee77:
I can update #285 to also drop testing on Python 3.6 and we can support only 3.7, 3.8 and 3.9 for an upcoming release?

[...]

PR #323, opened 2021-12-09
Fixed setup.py and working CI builds

### LINKED ISSUE #322: can not install in archlinux (manjaro) (PY3.9)  (cython3)
I can not install via pip in manjaro :

```
> python -V   -> Python 3.9.7
> pip -V   -> pip 21.3.1
> uname -a -> Linux 5.15.2-2-MANJARO #1 SMP PREEMPT  x86_64 GNU/Linux
> cython -> Cython version 3.0.0a
```

```

 ERROR: Command errored out with exit status 1:
   command: /usr/bin/python -u -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-1ncfokx8/pyfftw_e6929d2d84934a12acde5629427f85a1/setup.py'"'"'; __file__='"'"'/tmp/pip-install-1ncfokx8/pyfftw_e6929d2d84934a12acde5629427f85a1/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' bdist_wheel -d /tmp/pip-wheel-ieflly8y
       cwd: /tmp/pip-install-1ncfokx8/pyfftw_e6929d2d84934a12acde5629427f85a1/
  Complete output (795 lines):
  running bdist_wheel
  running build
  running build_py
  creating build
  creating build/lib.linux-x86_64-3.9
  creating build/lib.linux-x86_64-3.9/pyfftw
  copying pyfftw/config.py -> build/lib.linux-x86_64-3.9/pyfftw
  copying pyfftw/_version.py -> build/lib.linux-x86_64-3.9/pyfftw
  copying pyfftw/__init__.py -> build/lib.linux-x86_64-3.9/pyfftw
  creating build/lib.linux-x86_64-3.9/pyfftw/builders
  copying pyfftw/builders/builders.py -> build/lib.linux-x86_64-3.9/pyfftw/builders
  copying pyfftw/builders/_utils.py -> build/lib.linux-x86_64-3.9/pyfftw/builders
  copying pyfftw/builders/__init__.py -> build/lib.linux-x86_64-3.9/pyfftw/builders
  creating build/lib.linux-x86_64-3.9/pyfftw/interfaces
  copying pyfftw/interfaces/scipy_fftpack.py -> build/lib.linux-x86_64-3.9/pyfftw/interfaces
  copying pyfftw/interfaces/scipy_fft.py -> build/lib.linux-x86_64-3.9/pyfftw/interfaces
  copying pyfftw/interfaces/numpy_fft.py -> build/lib.linux-x86_64-3.9/pyfftw/interfaces
  copying pyfftw/interfaces/dask_fft.py -> build/lib.linux-x86_64-3.9/pyfftw/interfaces
  copying pyfftw/interfaces/cache.py -> build/lib.linux-x86_64-3.9/pyfftw/interfaces
  copying pyfftw/interfaces/_utils.py -> build/lib.linux-x86_64-3.9/pyfftw/interfaces
  copying pyfftw/interfaces/__init__.py -> build/lib.linux-x86_64-3.9/pyfftw/interfaces
  UPDATING build/lib.linux-x86_64-3.9/pyfftw/_version.py
  set build/lib.linux-x86_64-3.9/pyfftw/_version.py to '0.12.0'
  running build_ext
  cythoning pyfftw/pyfftw.pyx to pyfftw/pyfftw.c
  /usr/lib/python3.9/site-packages/Cython/Compiler/Main.py:346: FutureWarning: Cython directive 'language_level' not set, using '3str' for now (Py3). This has changed from earlier releases! File: /tmp/pip-install-1ncfokx8/pyfftw_e6929d2d84934a12acde5629427f85a1/pyfftw/pyfftw.pxd
    tree = Parsing.p_module(s, pxd, full_module_name)
  
  Error compiling Cython file:
  ------------------------------------------------------------
  ...
  # the types supported in this build
  _supported_types = []
  _supported_nptypes_complex = []
  _supported_nptypes_real = []
  
  IF HAVE_SINGLE:
    ^
  ------------------------------------------------------------
  
  pyfftw/pyfftw.pyx:72:3: Compile-time name 'HAVE_SINGLE' not defined
  
  Error compiling Cython file:
  ------------------------------------------------------------
  ...
  
  IF HAVE_SINGLE:
      _supported_types.append('32')
      _supported_nptypes_complex.append(np.complex64)
      _supported_nptypes_real.append(np.float32)
  IF HAVE_DOUBLE:
    ^
  ------------------------------------------------------------
  
  pyfftw/pyfftw.pyx:76:3: Compile-time name 'HAVE_DOUBLE' not defined
  
  Error compiling Cython file:
  ------------------------------------------------------------
  ...
      _supported_nptypes_real.append(np.float32)
  IF HAVE_DOUBLE:
      _supported_types.append('64')
      _supported_nptypes_complex.append(np.complex128)
      _supported_nptypes_real.append(np.float64)
  IF HAVE_LONG:
    ^
  ------------------------------------------------------------
  
  pyfftw/pyfftw.pyx:80:3: Compile-time name 'HAVE_LONG' not defined
  
  Error compiling Cython file:
  ------------------------------------------------------------
  ...
  IF HAVE_LONG:
      _supported_types.append('ld')
      _supported_nptypes_complex.append(np.clongdouble)
      _supported_nptypes_real.append(np.longdouble)
  
  IF (HAVE_SINGLE_OMP or HAVE_DOUBLE_OMP or HAVE_LONG_OMP):
     ^
  ------------------------------------------------------------
  
  pyfftw/pyfftw.pyx:85:4: Compile-time name 'HAVE_SINGLE_OMP' not defined
```

JeppeKlitgaard:
I spent much of today fiddling around with trying to get `pyfftw` working on Windows with Python 3.9 and `cibuildwheel` build wheels based on @grlee77's work over in his fork.

Since I was trying to get it to work on Python 3.9+, the changes ended up becoming a bit of a monolithic beast.

I would like some input from @grlee77 in particular before removing the [WIP], although it should mostly be working at this stage. In particular, I am curious why we are pinning numpy versions so stringently in `pyproject.toml`? I suspect this might be why we are unable to use wheels on the Linux CI-builds. Further, would these not lead to conflicts with a lot of projects that require more recent NumPy versions?

Brief overview:
- CI works for Windows, Linux, MacOS
- Compatible with all Python versions that have not reached EOL (3.7, 3.8, 3.9, 3.10)
- Wheels for all Python versions that are not dead for most architectures and operating systems
- `setup.py` updated with fixes for Windows and Python 3.9+
- We use only `manylinux2014`, as `manylinux2010` has reached EOL. Largest impact of this seems to be lack of support for distributed `pip` in Ubuntu 18.04, but people should not be running that old of a `pip` anyway.
- Runs `setup(...)` twice to establish which FFTW libraries are available and then inject into Cython compilation (fixes #322 and is neccessary)

# Wheel availability through CI

| Python version | Windows (32 bit) | Windows (64 bit) | Windows ARM (64 bit) | MacOS | MacOS ARM | Linux (32 bit) | Linux (64 bit) | Linux ARM (64 bit) |
| -------------- |:----------------:|:----------------:|:--------------------:|:-----:|:---------:|:--------------:|:--------------:|:------------------:|
| > 3.7          |        ❌        |        ❌        |          ❌          |  ❌   |    ❌     |       ❌       |       ❌       |         ❌          |
| 3.7            |        ❌        |        ✔         |          ❌          |   ✔   |    ❌     |       ✔        |       ✔        |         ✔          |
| 3.8            |        ❌        |        ✔         |          ❌          |   ✔   |    ❌     |       ✔        |       ✔        |         ✔          |
| 3.9            |        ❌        |        ✔         |          ❌          |   ✔   |    ❌     |       ✔        |       ✔        |         ✔          |
| 3.10           |        ❌        |        ✔         |          ❌          |   ✔   |    ❌     |       ❌        |       ✔        |         ✔          |
# Todo

These may belong in a separate PR, but build instructions for Windows given in README do not work. An additional step to translate the `.def` files into `.lib` are needed when using `msvc`.

# Issues affected

Should close the following issues:
- #301 
- #309 
- #266 
- #322 

Further, by enabling us to do a release, we can close the following issues:

- #317 
- #312 (by having latest tag be an ancestor)
- #252 (I think)
- #234 
- #144 

---

I think it would be nice to have this merged sooner rather than later as there seems to be a fair few people struggling with pyfftw on modern Pythons.


grlee77:
Hi @JeppeKlitgaard, thanks for resuming this work.

Have you been testing the wheel builds via GitHub actions yet? A reasonable way to do so is to enable GitHub Actions in your fork of the repository and then create a tag that has a name starting with `buildwheels` and push it to your fork. That should trigger the action to run. You can find it under the "Actions" tab and see if it succeeds.

I am happy to review again once you think things are in a reasonable state. I learned more about `cibuildwheel` in recent weeks while working on things for scikit-image.


> In particular, I am curious why we are pinning numpy versions so stringently in pyproject.toml? I suspect this might be why we are unable to use wheels on the Linux CI-builds. Further, would these not lead to conflicts with a lot of projects that require more recent NumPy versions?

The versions in `pyproject.toml` do not limit users of the library to those NumPy versions. It just specifies which version should be used during the build of a package. These are basically the oldest NumPy versions that meet our minimum requirement and have wheels available on PyPI. For example, there are no Python 3.10 wheels for NumPy < 1.21. We want to compile with as old of a NumPy as possible because newer versions will work with software compiled with an older version of NumPy, but the reverse is not true. If we compile with recent NumPy, we will likely prevent users with older NumPy from being able to use it.

Often we can follow what SciPy is using: https://github.com/scipy/scipy/blob/master/pyproject.toml (they have already dropped Python 3.7, though, so it won't be quite the same).


One change I would recommend is putting an upper bound on setuptools in that file. We might want to pin to `setuptools<=59.4.0` (the current release). This is because of some [planned changes across setuptools and numpy.distutils](https://mail.python.org/archives/list/numpy-discussion@python.org/thread/JIN5YSD7CAJACRGXMD3S4S5B5MUDWZVL/) that could break builds going forward.

grlee77:
I don't think there should be issues finding wheels for Python 3.10 anymore at this point. When I first started working on this, some of our dependencies didn't have these yet, but at this point they do.

JeppeKlitgaard:
Thank you so much for having a look.

I have everything working except for the Linux aarch64 builds, at which point it should be ready for you to have a look. I'll likely also remove the `appveyor` and `travis` configurations before removing `[WIP]`.

> The versions in `pyproject.toml` do not limit users of the library to those NumPy versions. It just specifies which version should be used during the build of a package. These are basically the oldest NumPy versions that meet our minimum requirement and have wheels available on PyPI. For example, there are no Python 3.10 wheels for NumPy < 1.21. We want to compile with as old of a NumPy as possible because newer versions will work with software compiled with an older version of NumPy, but the reverse is not true. If we compile with recent NumPy, we will likely prevent users with older NumPy from being able to use it.

That makes perfect sense - thank you for bearing with me!

> One change I would recommend is putting an upper bound on setuptools in that file. We might want to pin to `setuptools<=59.4.0` (the current release). This is because of some [planned changes across setuptools and numpy.distutils](https://mail.python.org/archives/list/numpy-discussion@python.org/thread/JIN5YSD7CAJACRGXMD3S4S5B5MUDWZVL/) that could break builds going forward.

I will add this here then.

---

I _think_ I got the `aarch64` working now.

The `win64` and `macOS` is all done.

.github/workflows/wheel_tests_and_release.yml
```diff
@@ -0,0 +1,204 @@
...
+
+      - name: Install cibuildwheel
+        run: |
+          python -m pip install cibuildwheel
+
+      - name: Build the wheel
+        run: |
+          python -m cibuildwheel --output-dir dist
+        env:
+          CIBW_BUILD: ${{ matrix.cibw_python }}
```
grlee77, line 52:
```suggestion
          CIBW_SKIP: "*-musllinux_*"
          # skip aarch64 tests which often exceed the 6 hour time limit
          CIBW_TEST_SKIP: "*_aarch64"
          CIBW_BUILD: ${{ matrix.cibw_python }}
```

I think I did have problems for `aarch64` and `musllinux` builds and dropped them from scikit-image. I think we can skip `musllinux` as pretty niche. If someone needs it we can add it later.

Not sure if you want the `CIBW_TEST_SKIP` here or not. On scikit-image the test suite could not reliably complete within the time limit for aarch64 so we had to skip it here.

JeppeKlitgaard: I agree that `musllinux` is definitely niche, but I would really like to get `aarch64` working if possible. Compiling from source is no fun on a Raspberry Pi!

grlee77: That variable is only to skip the **tests** on aarch64. Definitely agree that we should still build those wheels. I couldn't remember how long this takes for PyFFTW, but for scikit-image it only sometimes would complete within the 6 hour timeout window on GitHub Actions. If tests are completing within the time limit, then let's just not skip them

grlee77:
Also, in terms of version support, there are the [recommendations from NumPy's NEP-29](https://numpy.org/neps/nep-0029-deprecation_policy.html#drop-schedule) which a number of other scientific projects are following. The table there indicates dropping support for NumPy 1.18 for releases after Dec. 22nd and Python 3.7 after Dec. 26th, but it is not a problem if we want to support them a bit longer than that.

JeppeKlitgaard:
> Also, in terms of version support, there are the [recommendations from NumPy's NEP-29](https://numpy.org/neps/nep-0029-deprecation_policy.html#drop-schedule) which a number of other scientific projects are following. The table there indicates dropping support for NumPy 1.18 for releases after Dec. 22nd and Python 3.7 after Dec. 26th, but it is not a problem if we want to support them a bit longer than that.

I am all for dropping old snakes - I think 3.7 is a pain point for many people due to Ubuntu 18.04 LTS though. I don't have a strong opinion on whether to follow the scientific python community deprecation schedule or the PSF schedule.

.github/workflows/wheel_tests_and_release.yml
JeppeKlitgaard: I have removed `musllinux` from the CI builds.

I think the tests run quite quickly, since they are just: `python -c "import pyfftw; import numpy as np; from pyfftw.interfaces import numpy_fft as fft; np.testing.assert_allclose(fft.fft(np.zeros(8)), np.zeros(8));"` (i.e. not the full or even 'quick' test suite).

JeppeKlitgaard:
I think I managed to find a combination of the manylinux build images that allows the use of `numpy` wheels. This will change in the future though, as `numpy` likes to have slightly more up-to-date manylinux images than strictly necessary.

Could you have a look at the artifacts @grlee77? It would be really nice to have the wheels (and misc. smaller fixes in `setup.py`)  along with the r2r fix out in a release.

grlee77:
> I think I managed to find a combination of the manylinux build images that allows the use of numpy wheels. This will change in the future though, as numpy likes to have slightly more up-to-date manylinux images than strictly necessary.

Ah, I remember that I got this to work on scikit-image by telling pip to "prefer binary" so that it will look for the most recent release with a wheel in preference to trying to compile the most recent version from source if it is missing one. You can add the option via an environment variable that `cibuildwheel` will find.

`CIBW_ENVIRONMENT: PIP_PREFER_BINARY=1`

grlee77:
I tried out the linux x86_64 Python 3.8 artifact and it is working well (test suite is still running, but no failures so far). I also tried some basic usage of the Windows AMD64 Python 3.9 wheel and that worked fine as well. I can potentially also try on an x86_64 Mac, but I don't have newer Apple M1 hardware.

I am out of office in the morning (US Eastern) tomorrow, but should be able to take another look here in the afternoon. It is great to see this coming together!

JeppeKlitgaard:
I'll try the `PIP_PREFER_BINARY`. As far as I can tell the only thing that is still failing is the testing run of the Linux 32-bit build for python 3.10. At a cursory glance it looks like it is trying to compile scipy but lacks OpenBLAS libraries. NumPy and SciPy are dropping 32-bit Linux wheels for 3.10 and beyond, I believe.

Options:
1. Offer cp310-i686 manylinux wheel but skip testing
2. Don't offer cp310-i686 manylinux wheel - anyone needing it would likely need to compile SciPy and NumPy from scratch and thus would be able to figure out how to compile pyfftw as well
3. Try to get SciPy to compile (will require very long run)

I think only 1 and 2 are valid options, with 2 being preferable to reduce maintenance workload. Let me know which you prefer and I'll finish it up.

Thank you so much for testing the artifacts - much appreciated!

hgomersall:
This is all great, thanks both!

grlee77:
I would follow NumPy's example and drop 32-bit for Python 3.10

JeppeKlitgaard:
I deleted `appveyor` configs in 355e79f - was this wrong? It seems to be running tests for PR's or something.

I had assumed this was a legacy service, but maybe it isn't?

JeppeKlitgaard:
All done - see last CI builds: https://github.com/JeppeKlitgaard/pyFFTW/runs/4485860511

I added a description of wheels that CI builds to the README to make it clearer when people should expect `pip` to be able to find binaries and when they would need to build from source themselves.

Should be ready to merge assuming appveyor/travis CI setups where indeed legacy and should be removed (see comment above)

hgomersall:
I'd prefer to defer to @grlee77 on this one, but I'm happy in principle once he is.

hgomersall:
Do we need to rip out the appveyor stuff?

hgomersall:
Or fix them?

JeppeKlitgaard:
I am not sure what the appveyor stuff did - I assumed it was a previous CI configuration for building wheels, but maybe not?

As far as I can tell, it was working. I haven't used appveyor before. I am happy to restore it if it was actually doing something.
Looking at it, it seems like it built Windows wheels and uploaded them to pypi and (now defunct) bintray?

It did also run `python setup.py test`, which we currently do not do in this. It seems excessive to do a full test for all 19 wheels.

Could implement the full test suite on some of the wheels though? Probably best done in a separate PR.

hgomersall:
Yeah, it mostly created windows wheels and ran the tests on those. Is that done with the gh runner now?

JeppeKlitgaard:
I think Travis did the same, but for Unix?

Both of these solutions seem like they are better replaced by the CI builds of this PR.

Given that @grlee77 had good success with the wheels so far, I suggest we merge this now/soon and then I can have a go at seeing whether expanding the workflow to run the 'quick tests' (rather than the 'can it import+' test that it is currently doing) for all wheels is viable or whether we should only do some wheels

JeppeKlitgaard:
> Yeah, it mostly created windows wheels and ran the tests on those. Is that done with the gh runner now?

Yea, we build AMD64 wheels for Windows for python 3.7 - 3.10 (4 wheels). 32 bit Windows could probably be supported too, with a bit more effort. Similarly for 64 bit ARM (only modern pythons only)

.github/workflows/wheel_tests_and_release.yml
```diff
@@ -0,0 +1,193 @@
...
+          PYFFTW_VERSION=$(git describe --tags)
+          python setup.py sdist
+          ls -la ${{ github.workspace }}/dist
+          # We prefer to release wheels before source because otherwise there is a
+          # small window during which users who pip install pyfftw will require compilation.
+          twine upload ${{ github.workspace }}/dist/*.whl
+          twine upload ${{ github.workspace }}/dist/pyfftw-${PYFFTW_VERSION:1}.tar.gz
+        env:
+          TWINE_USERNAME: ${{ secrets.TWINE_USERNAME }}
+          TWINE_PASSWORD: ${{ secrets.TWINE_PASSWORD }}
```
grlee77, line 187: @hgomersall, can you add `TWINE_USERNAME` and `TWINE_PASSWORD` as secrets on this repository? The process to add them is going to the "Settings" for the repository and selecting "Secrets" from the menu at the left. I do not have the permissions required to do it.

If they are not defined, automated deployment will fail, but we can still manually download the artifacts and upload the release to PyPI via `twine`.

.github/workflows/wheel_tests_and_release.yml
```diff
@@ -0,0 +1,193 @@
...
+
+      - name: Github release
+        uses: softprops/action-gh-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
grlee77, line 192: @hgomersall: similarly to above, this step requires a `GITHUB_TOKEN` secret to be defined for the repository. (This step is publishing as a release on the GitHub releases page)

.travis.yml
```diff
@@ -1,72 +0,0 @@
-language: python
```
grlee77, line 1: It is useful to keep this file for now, but we can delete the `before_deploy` and `deploy` sections. This runs tests on the CI services for new PRs.

Ideally, we could port these tests over to a GitHub Actions workflow named something like `.github/workflows/tests.yml`. About a year or so ago, TravisCI started more strictly limiting the amount of free minutes for open source projects, but given the low activity on this repository, it may still work okay for us.

README.md
```diff
@@ -102,6 +102,21 @@ where the version and the revision hash are set accordingly.
 
 Read on if you do want to build from source...
 
+## Wheels
```
grlee77, line 105: Nice Table!

appveyor.yml
```diff
@@ -1,105 +0,0 @@
-environment:
```
grlee77, line 1: Again, we can keep this for now until/unless we have a replacement. This will run tests on every PR commit, whereas the wheel build script currently only runs on specific tags.

certainly fine to remove deployment steps from here, though.

grlee77: will have to keep `appveyor/setup_fftw_dlls.cmd` and `appveyor/run_with_env.cmd` files as well, but it does seem safe to delete the `deploy_script` section and `deploy.ps1`.

pyproject.toml
```diff
@@ -1,12 +1,173 @@
...
+    [tool.cibuildwheel.macos]
+    archs = [
+        "x86_64",
+
+        # Unfortunately we don't have Apple Silicon support yet
+        # TODO: Fix this
+        # Current problem seems to be that installed libfftw3 does not provide arm64
+        # symbols
+        # "universal2",
+        # "arm64",
```
grlee77, line 27: Even though we don't provide this here at the moment, we can likely provide arm64 support via conda-forge.

pyproject.toml
```diff
@@ -1,12 +1,173 @@
...
+    ]
+
+        [tool.cibuildwheel.macos.environment]
+        CC = "/usr/bin/clang"
+        CXX = "/usr/bin/clang++"
+        CFLAGS = "-Wno-implicit-function-declaration -I/usr/local/opt/fftw/include"
+        LDFLAGS = "-Wl,-rpath,/usr/local/opt/fftw/lib -L/usr/local/opt/fftw/lib"
```
grlee77, line 34: Not sure if it is worth the effort for PyFFTW as there is a lot less Cython code than in scikit-image, but it is possible to shrink the size of the binary wheels by stripping debug symbols. For clan this would be adding `-Wl,-S ` here.

```suggestion
        LDFLAGS = "-Wl,-S -Wl,-rpath,/usr/local/opt/fftw/lib -L/usr/local/opt/fftw/lib"
```
At least for scikit-image this shrinks wheel size by more than a factor of two.

I ended up doing that for Linux as well via gcc's ``-Wl,--strip-debug`, but I struggled a bit getting the environment variable to be found. I didn't get it to work with LDFLAGS, but it worked when defining a [custom environment variable](https://github.com/scikit-image/scikit-image/blob/1a14bb32b5a6718870dc0b3df186b7ed3018f337/.github/workflows/wheel_tests_and_release.yml#L51-L52) that was [checked in setup.py](https://github.com/scikit-image/scikit-image/blob/1a14bb32b5a6718870dc0b3df186b7ed3018f337/setup.py#L119-L120).

grlee77: Just checked the wheels and they are all around 1.5-3 MB so this is probably not too high of a priority. For scikit-image we had sizes >30MB and we able to shrink to around 13MB.

pyproject.toml
```diff
@@ -1,12 +1,173 @@
...
+
+    # The best choice of manylinux is determined by the following:
+    # manylinux version for numpy version for given python version
+
+    # Use manylinux1 for Python3.7, as <3.7.3 does not initially work with manylinux2010
+    [[tool.cibuildwheel.overrides]]
+    select = "cp37-manylinux*"
+
+    manylinux-x86_64-image = "manylinux1"
+    manylinux-i686-image = "manylinux1"
```
grlee77, line 72: I think some packages went ahead and also bumped these to manylinux2010, but it is true that this doesn't work for the version of `pip` bundled with Python 3.7. seems fine to leave it at manylinux1 in this Python 3.7 case.

setup.py
```diff
@@ -33,27 +33,21 @@
...
-    from distutils.core import setup, Command
-    from distutils.command.build_ext import build_ext
-    using_setuptools = False
-
-from distutils import log
-from distutils.ccompiler import get_default_compiler, new_compiler
+from setuptools import setup, Command
+from setuptools.command.build_ext import build_ext
+
 from distutils.errors import CompileError, LinkError
```
grlee77, line 39: Let's leave it to a future PR, but we need to eventually move away from `distutils` going forward. It is deprecated in Python 3.10 and will be removed in Python 3.12 (if I recall correctly).

.travis.yml
grlee77: requires also keeping `travis/partial.sh`

grlee77 (changes requested):
Thanks a lot @JeppeKlitgaard!

This looks good to me overall, but I think we need three things before merging:

1.) let's restore the Travis and Appveyor scripts (although removing the deployment steps is fine)
2.) @hgomersall should add the secrets to the repository (see in-line code comments)
3.) we should also test the changes in `setup.py` with the conda recipe at https://github.com/conda-forge/pyfftw-feedstock and make sure the recipe still runs. Superficially these changes look okay, but it would be good to test further in that scenario.

.github/workflows/wheel_tests_and_release.yml
hgomersall: Is this the PyPI API tokens?

hgomersall: It looks like there is a TestPyPI as well as a PyPI. I think the name should be different here (not twine, but TestPyPI, if that's what is needed).

README.md
hgomersall: I agree. There's some legacy stuff above that table in the installation section that needs pruning. It would be great if you could do that bit (including it seems, the stuff on bintray). Your table is much better than most of what else is there. Out of interest, do we still build on every commit?

hgomersall:
@grlee77 why do we need travis and appveyor? Is that not handled by gh actions now?

.github/workflows/wheel_tests_and_release.yml
hgomersall: Hmm, I'd rather not do this as my personal account. Is there a way for the project to publish itself? Can this be done with deploy keys?

README.md
JeppeKlitgaard: Currently these are only built on: releases, commits in `cibuildwheel` branches.

pyproject.toml
JeppeKlitgaard: I am not at all familiar with building conda packages - is there anything that I need to do for conda support?

pyproject.toml
JeppeKlitgaard: I think these are best kept as manylinux1 so it does actually work for all Python 3.7 installations. Support for manylinux1 is disappearing soon (in pypa/manylinux) and once Python3.7 support can be dropped, we can start dropping some of these overrides.

In Python 3.10+ there should be no need for overrides at all.

setup.py
JeppeKlitgaard: The documentation for `setuptools` is only marginally better than that of `distutils`. I didn't find an easy guide to update this to a future-compatible `setup.py`. I hope that once Python 3.12 approaches, the documentation on how to switch away from `distutils` will be a lot better.

pyproject.toml
JeppeKlitgaard: Is this meant to be `-Wl,-S -Wl` and not `-S -Wl`? Would those not be the same?

.github/workflows/wheel_tests_and_release.yml
grlee77: Actually, I think I was mistaken on this one and it may already be defined for the repo by default. It should not be your personal GitHub access token, but should be specific to only this repository. There are some details on setting more [granular permissions in this post](https://dev.to/github/the-githubtoken-in-github-actions-how-it-works-change-permissions-customizations-3cgp), but I haven't tried it and don't know what the minimum are that would be needed to publish a release.

pyproject.toml
grlee77: No, nothing specifically. I can try running the recipe later with the changes proposed here just to make sure it still works. I just wanted to double check that the changes to `setup.py` are not going to break the builds there.

pyproject.toml
grlee77: No, the `-Wl,-S` must be specified together as suggested here. Options to be passed on from the compiler to the linker are prefixed with "-Wl,". There is a lot more info [here](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/developer_guide/gcc-using-libraries) under the section "Passing the -Wl option" (That is for gcc but I think it is the same for clang). From that description it could also be concatenated into a single `-Wl` with 3 options afterward:  `-Wl,-S,-rpath,/usr/local/opt/fftw/lib`.

On second look, I would have thought it would have an `=` instead of `,` between `-rpath` and `/usr/local/opt/fftw/lib`, but [answers here state that either form is fine](https://stackoverflow.com/questions/6562403/i-dont-understand-wl-rpath-wl).

JeppeKlitgaard:
@hgomersall @grlee77 

I am trying to implement the testing in GitHub Actions as I find the idea of having 3 CI services set up a little daunting and will likely be very hard to maintain.

I think Linux 32-bit was never tested using Travis and I am now running into a failed test that I believe stems from a bug. 

An example of this can be seen in: https://github.com/JeppeKlitgaard/pyFFTW/runs/4511210933?check_suite_focus=true

```
=================================== FAILURES ===================================
  ___________________ ByteAlignTest.test_byte_align_set_dtype ____________________
  
  self = <tests.test_pyfftw_nbyte_align.ByteAlignTest testMethod=test_byte_align_set_dtype>
  
      def test_byte_align_set_dtype(self):
          shape = (10,10)
          a = numpy.int16(numpy.random.randn(*shape)*16000)
  Error: Command ['sh', '-c', 'pytest --import-mode=append /project'] failed with code 1. 
          b = numpy.float64(numpy.random.randn(*shape))
  
          c = numpy.int8(numpy.random.randn(*shape)*255)
          # Test a few alignments
          for n in [None, 3, 7, 9, 16, 24, 23, 63, 64]:
              expected_alignment = get_expected_alignment(n)
              d = byte_align(a, dtype='float32', n=n)
  >           self.assertTrue(d.ctypes.data % expected_alignment == 0)
  E           AssertionError: False is not true
  
  /project/tests/test_pyfftw_nbyte_align.py:322: AssertionError
 
  =========================== short test summary info ============================
  FAILED ../project/tests/test_pyfftw_nbyte_align.py::ByteAlignTest::test_byte_align_set_dtype
  ===== 1 failed, 1790 passed, 8 skipped, 649 warnings in 629.92s (0:10:29) ======
```

Notably this only fails on 32-bit Linux and not on 64-bit Linux, which is why I think this could a hitherto undiscovered bug. Is there a chance either of you could have a look?

JeppeKlitgaard:
Current progress:

Windows: Working fully w/ testing
Linux: Fails on 32-bit on a single test (see above)
MacOS: For some reason the wheel is not being properly installed in the test venv by `cibuildwheel`. I don't have access to a Mac, which makes debugging this a little harder. Specifically, it errors with "ModuleNotFoundError: pyfftw.pyfftw`

The generated wheel seems to be fine on the surface when inspecting with `wheel2json`.

I propose a temporary solution:
- Skip tests on Linux 32-bit, but create issue detailing the failing test. I believe this is unrelated to the CI workflow and is an actual bug that has not previously been discovered
- Skip tests on MacOS. I believe the wheel is fine, but testing setup is funky. Create an issue detailing this

Even with those temporary overrides, this PR still supersedes the testing implemented using Travis and Appveyor previously. (Namely: Linux 64 bit and Windows).

Thoughts?

JeppeKlitgaard:
See #325 #324 

Note:
- `test` directory changed to `tests` to avoid namespace conflict with the CPython module `test`. See: https://docs.python.org/3/library/test.html
- `appveyor` and `travis` are removed in favour of testing through `GitHub Actions`
- Build+Test has been enabled on PR's and commits to `master`
- Build+Test+Deploy is run on release tags
- Tests disabled for Linux 32-bit builds (See: #325)
- Tests disabled for MacOS builds (See: #324)

pyproject.toml
JeppeKlitgaard: Ah, that makes sense - thank you!

I changed it to `"-Wl,-S,-rpath=/usr/local/opt/fftw/lib -L/usr/local/opt/fftw/lib"` which I think is the most reading-friendly version.

Leaving this open due to:

> I ended up doing that for Linux as well via gcc's ``-Wl,--strip-debug`, but I struggled a bit getting the environment variable to be found. I didn't get it to work with LDFLAGS, but it worked when defining a custom environment variable that was checked in setup.py.

pyproject.toml
JeppeKlitgaard: Perfect, thank you. I wouldn't think the `setup.py` changes would be breaking, but would be good to check.

grlee77:
Testing from this branch on conda-forge here: https://github.com/conda-forge/pyfftw-feedstock/pull/46

JeppeKlitgaard:
`conda-forge` seems to be passing.

Unfortunately when running the tests on `aarch64` (ARM 64) Linux, 21 tests fail.

Additionally running the full test-suite on those seems to be very lengthy, sometimes running into the 6h limit on GitHub Actions.

I will disable testing on `aarch64` and create an issue like I did for Linux i686. 

If we ever add `aarch64` testing it might be worthwhile to use `pytest` markers to make a `quicktest` testing suite that only runs some tests.

appveyor.yml
JeppeKlitgaard: N/A after GHA

.travis.yml
JeppeKlitgaard: N/A after GHA

JeppeKlitgaard:
Builds and tests are now passing. (Tests disabled for MacOS, Linux aarch64, Linux i686 - see #324, #325, #326)

Remaining:

- [ ] Strip-debug Linux
- [ ] Credentials (@hgomersall)

hgomersall:
FYI I think there is a quick test...

hgomersall:
> Builds and tests are now passing. (Tests disabled for MacOS, Linux aarch64, Linux i686 - see #324, #325, #326)
> 
> Remaining:
> 
>     * [ ]  Strip-debug Linux
> 
>     * [ ]  Credentials (@hgomersall)

I'm not sure what's needed for github deployment. Are these personal credentials?

hgomersall:
Quick test: https://github.com/pyFFTW/pyFFTW/blob/d74d032561f0ba712c5d99c23bbfb9e336c3d3ee/setup.py#L693

JeppeKlitgaard:
@hgomersall I think just the PYPI credentials are needed. The GitHub credentials are automatically replaced. (See: https://github.com/pyFFTW/pyFFTW/pull/323#discussion_r767800352)

Relevant GH documentation: https://docs.github.com/en/actions/security-guides/encrypted-secrets

pyproject.toml
JeppeKlitgaard: @grlee77 interestingly I only see a reduction of ≈ 10 KB out of the ≈ 3 MB wheel on MacOS.

I will leave in the MacOS flags, but won't bother with Linux since the reduction is negligible.

JeppeKlitgaard:
I changed the CI config a bit to fully parallelize all builds, which should speed up builds substantially.

From my point of view, this is mergeable as is. It supersedes the previous Travis+Appveyor setup and builds on all supported Pythons.

@grlee77:

- Could you give this another quick once-over and maybe test a MacOS wheel if you have the available hardware. (See: #324)

@hgomersall:

- This is safely mergeable without credentials, but @somebody below is not

@somebody:

- Once merged, could you make a new release (just using a tag) and push it?
- Verify that CI builds+testing works
- Verify that wheels are deployed to Pypi

# Going forward

Once/if some of #324, #325, #326 are fixed we should reenable more tests. Note that aarch64 tests are slow

hgomersall:
Are we able to test the credentials are correct prior to merging?

JeppeKlitgaard:
Not really. You could test using the TestPyPi credentials, but there wouldn't be much point in that I think.

If the credentials are wrong the deployment will simply fail and we can fix the credentials

.github/workflows/wheel_tests_and_release.yml
JeppeKlitgaard: Resolving since `GITHUB_` secrets are automatically set.

README.md
JeppeKlitgaard: Pruning done

grlee77:
All looks well on conda-forge! (tried versions of the recipe both with and without static linking of FFTW)

grlee77:
> Not really, I think. You could test using the TestPyPi credentials, but there wouldn't be much point in that I think.

Agreed. If you are worried about it we can tag a release candidate first and just update and try again. 

Even if it fails there is always the fallback of:
1.) download the artifact and extract the wheels
2.) upload the extracted wheels to PyPI using twine

I had to do the manual step for scikit-image 0.19.0 because mid-upload PyPI gave an error about exceeding a space quota (there is a 10GB per project total limit and 100 MB per wheel limit by default)

.github/workflows/wheel_tests_and_release.yml
```diff
@@ -0,0 +1,213 @@
+name: Build Wheels and Release
+
+on:
+  push:
+    branches:
+      - cibuildwheel
+      - master
```
grlee77, line 7:
```suggestion
      - master
```
might just remove cibuildwheel once you are done testing since we probably don't want to keep a cibuildwheel branch on the main repo. we can always still push a tag starting with `buildwheels` to test things out there.

.github/workflows/wheel_tests_and_release.yml
```diff
@@ -0,0 +1,213 @@
...
+      - name: Build the wheel
+        run: |
+          python -m cibuildwheel --output-dir dist
+        env:
+          CIBW_BUILD: ${{ matrix.cibw_python }}-${{ matrix.cibw_libc }}_${{ matrix.cibw_arch }}
+
+      - uses: actions/upload-artifact@v2
+        with:
+          name: wheels
+          path: ./dist/*.whl
```
grlee77, line 62: can we have some type of if clause to disable this on pull_requests? Otherwise each time a user makes a commit in a PR we will generate and store wheel artifacts, which seems wasteful and not very useful (don't recall offhand how long GitHub retains the artifacts).

grlee77 (commented):
Do we have to merge this before we will be able to see the CI results in the PR? Right now there is a residual Appveyor showing up only, but normally would expect to see the results of the GitHub test cases in the bottom of the PR.

grlee77:
> Could you give this another quick once-over and maybe test a MacOS wheel if you have the available hardware.


I ran the macosx x86_64 wheel for Python 3.9 and it seems okay (using the most recent wheel artifact from your fork: https://github.com/JeppeKlitgaard/pyFFTW/actions/runs/1578558142).

.github/workflows/wheel_tests_and_release.yml
JeppeKlitgaard: Done, I think. Bit hard to test for now. I believe retention is 90 days?

JeppeKlitgaard:
> Actually, let me ammend my prior claim. I do see the errors as reported in #324 if I try to run the pyFFTW test suite using the version installed from the wheels. However, just directly running things like:
> 
> ```python
> import numpy as np
> import pyfftw.interfaces.numpy_fft as fft
> 
> fft.ifft(fft.fft(np.arange(16, dtype=float)))
> ```
> 
> works as expected

I think this must be some weirdness in the way that Python on MacOS resolves packages. I think it somehow finds a local package and not the installed one? If you have a MacOS computer you could potentially have a look in the `site-packages` directory where the `__init__.py` resides and see if the wheel is there or not.

Ideally this should be fixed, but it does not block the PR as this behaviour only affects MacOS testing which was not set up previously anyway. This should be dealt with in #324 after this is merged, in my opinion.

JeppeKlitgaard:
Remaining changes should not require any changes to files, just secrets and CI in GitHub interface.

grlee77:
>  Appveyor integration setup in GitHub interface and that overrides the GHA CI somehow

some other projects have both, so I doubt this is it

> GitHub uses master branch for the CI details on PRs

This seems most likely to me, but I am not sure

grlee77:
I ran the full test suite using the MacOS wheel and got a single test failure related to caching. It is likely a pre-existing issue since we do not seem to have been testing on mac previously:

<details>

```
=============================================================== FAILURES ================================================================
____________________________________________ CacheTest.test_objects_removed_after_keepalive _____________________________________________

self = <tests.test_pyfftw_interfaces_cache.CacheTest testMethod=test_objects_removed_after_keepalive>

    @unittest.skipIf(*miss('64'))
    def test_objects_removed_after_keepalive(self):
        _cache = interfaces.cache._Cache()
    
        key = 'the key'
    
        test_array = numpy.random.randn(16)
        obj = builders.fft(test_array)
        _cache.insert(obj, key)
    
        self.assertIs(_cache.lookup(key), obj)
    
        keepalive_time = _cache.keepalive_time
    
        if os.name == 'nt':
            # A hack to keep appveyor from falling over here. I suspect the
            # contention is too much to work properly. Either way, let's
            # assume it's a windows problem for now...
            time.sleep(keepalive_time * 8)
        else:
            # Relax a bit more otherwise
            time.sleep(keepalive_time * 4)
    
        _cache.insert(obj, key)
        old_keepalive_time = _cache.keepalive_time
        _cache.set_keepalive_time(old_keepalive_time * 4)
    
        self.assertIs(_cache.lookup(key), obj)
    
        time.sleep(old_keepalive_time * 3.5)
        # still should be there
        self.assertIs(_cache.lookup(key), obj)
    
        if os.name == 'nt':
            # As above, but with a bit longer
            time.sleep(old_keepalive_time * 16)
        else:
            time.sleep(old_keepalive_time * 8)
    
>       self.assertRaises(KeyError, _cache.lookup, key)
E       AssertionError: KeyError not raised by lookup

test_pyfftw_interfaces_cache.py:328: AssertionError
=========================================================== warnings summary ============================================================
test_r2r.py:125
  /Users/glee/Downloads/pyFFTW_wheels/pyFFTW/tests/test_r2r.py:125: PytestCollectionWarning: cannot collect test class 'TestRealTransform' because it has a __init__ constructor (from: tests/test_r2r.py)
    class TestRealTransform(object):

tests/test_pyfftw_dask_interface.py: 408 warnings
  /Users/glee/mambaforge/envs/py39/lib/python3.9/site-packages/dask/array/core.py:1503: FutureWarning: The `numpy.alltrue` function is not implemented by Dask array. You may want to use the da.map_blocks function or something similar to silence this warning. Your code may stop working in a future release.
    warnings.warn(

tests/test_pyfftw_dask_interface.py: 120 warnings
  /Users/glee/mambaforge/envs/py39/lib/python3.9/site-packages/pyfftw/builders/_utils.py:204: ComplexWarning: Casting complex values to real discards the imaginary part
    a = numpy.asarray(a, dtype=_rc_dtype_pairs[a.dtype.char])

tests/test_pyfftw_dask_interface.py: 76 warnings
  /Users/glee/mambaforge/envs/py39/lib/python3.9/site-packages/numpy/fft/_pocketfft.py:70: ComplexWarning: Casting complex values to real discards the imaginary part
    r = pfi.execute(a, is_real, is_forward, fct)

tests/test_pyfftw_dask_interface.py: 44 warnings
  /Users/glee/mambaforge/envs/py39/lib/python3.9/site-packages/numpy/fft/_pocketfft.py:73: ComplexWarning: Casting complex values to real discards the imaginary part
    r = pfi.execute(a, is_real, is_forward, fct)

-- Docs: https://docs.pytest.org/en/stable/warnings.html
======================================================== short test summary info ========================================================
FAILED test_pyfftw_interfaces_cache.py::CacheTest::test_objects_removed_after_keepalive - AssertionError: KeyError not raised by lookup
================================== 1 failed, 1789 passed, 9 skipped, 649 warnings in 557.76s (0:09:17) ==================================
```

<\details>

JeppeKlitgaard:
> I ran the full test suite using the MacOS wheel and got a single test failure related to caching. It is likely a pre-existing issue since we do not seem to have been testing on mac previously:

Yea, this seems unrelated to CI. If you make a new issue for this, could you link it to #324?

pyproject.toml
```diff
@@ -1,12 +1,188 @@
...
+    [tool.cibuildwheel.linux]
+
+    # See: https://github.com/pyFFTW/pyFFTW/issues/325
+    # See: https://github.com/pyFFTW/pyFFTW/issues/326
+    test-skip = [
+        "*i686",
+        "*aarch64",
+    ]
+
+    test-command = "ls -la; echo $PYTHONPATH; pytest -v --import-mode=append {project}/tests"
```
grlee77, line 52: Did you try the tests on MacOS using this form of `test-command`? It differs from the one up top in that it lists `{project}/tests` instead of `{project}`!

JeppeKlitgaard: Whoops - that was debugging and meant to have been removed. Good catch.

I believe I tried both versions and that `{project}/tests` should resolve to the same test-suite. Will have a look at this now.

JeppeKlitgaard: Running CI again now having removed that. If Linux builds fail then this was likely the issue. Thank you for catching this!

JeppeKlitgaard: This has been resolved. See #324, #327

JeppeKlitgaard:
Apologies, I think we managed to clutter the discussion a bit when running into other issues.

The reason we initially wanted to restore TravisCI and Appveyor was because back then this PR did not run the test suites. It now does. Thus both TravisCI and Appveyor can be (and have been) removed.

Summary:
- We no longer need TravisCI or Appveyor
- All needed building, testing, and deployment through CI is now handled using GitHub Actions
- PR is ready to merge
- Pypi credentials need to be added (See: https://github.com/pyFFTW/pyFFTW/pull/323#issuecomment-993477213)

See also: https://github.com/pyFFTW/pyFFTW/pull/323#issuecomment-993569326

README.md
```diff
@@ -194,6 +200,23 @@ Install FFTW from ports tree or ``pkg``:
 
 Please install all of them, if possible.
 
+## Testing
+
+Tests should be run using `pytest`. Install using:
```
hgomersall, line 205: Is this necessary? Does `python -m unittest discover` not work any more?

README.md
```diff
@@ -1,8 +1,8 @@
...
-| --- | --- | --- |
-| [![travis_ci](https://travis-ci.org/pyFFTW/pyFFTW.svg?branch=master)](https://travis-ci.org/pyFFTW/pyFFTW) | [![appveyor_ci](https://ci.appveyor.com/api/projects/status/uf854abck4x1qsjj/branch/master?svg=true)](https://ci.appveyor.com/project/hgomersall/pyfftw) | [![read_the_docs](https://readthedocs.org/projects/pyfftw/badge/?version=latest)](http://pyfftw.readthedocs.io/en/latest/?badge=latest) |
+| GitHub Actions                                                                                                  | Read the Docs                                                                                                                           |
+| --------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------- |
+| ![github_actions](https://img.shields.io/github/workflow/status/JeppeKlitgaard/pyFFTW/wheel_tests_and_release)  | [![read_the_docs](https://readthedocs.org/projects/pyfftw/badge/?version=latest)](http://pyfftw.readthedocs.io/en/latest/?badge=latest) |
```
hgomersall, line 5: Presumably this needs updating to be the correct githubs action badge? (i.e. not yours)

README.md
JeppeKlitgaard: One could possibly continue to just use `unittest`, but for CI I use `pytest` because it allows us to specifically target the generated `wheel` over the local (uncompiled) python module.

I could not get `unittest` discoverer to easily place the installed wheel folder before the local directory in `sys.path`.

No changes to the workflow have been made, so `python -m unittest discover` will continue to work as before.

README.md
JeppeKlitgaard: Yes, absolutely. I think this change is better made by you directly to `master` after this is merged, as the configuration we want should target `master` and it is not possible to really test before this is merged.

README.md
JeppeKlitgaard: Edit: this is not entirely true. The folder with the tests have been renamed from `test` to `tests` to avoid a namespace conflict with the built-in `test` module from CPython.

I assume `unittest` will also discover the test directory when it is named `tests`. I believe it is recommended for it to be named `tests` anyway for the reason mentioned above. This change was necessary to ensure the proper module resolution.

JeppeKlitgaard: Feel free to change this if you prefer the other method to be the documented approach. The underlying tests remain the same - the only difference is which test runner is used. `pytest` arguable is more easily configurable and is a common choice in my experience.

hgomersall:
Once @grlee77 is happy with this, we can merge and fix any remaining CI issues afterwards - there are too many potential issues that can be worked out until then.

grlee77 (approved):
Looks good to me now. Let's get this in! Thanks for all the work here @JeppeKlitgaard

hgomersall:
I concur. Great work, thanks!

PR #331, opened 2021-12-17
CI: Added the correct gh actions badge and tweaked the name of the workflow

hgomersall:


hgomersall:
At @grlee77 should be trivial to merge this to make the badge all pretty again!

grlee77 (approved):
Great. Just checked the link and it works

Issue #327, opened 2021-12-14
failing cache test on MacOS


grlee77:
While locally testing wheels from #323, I encountered a single test failure related to caching. It is likely a pre-existing issue since we do not seem to have been testing on mac previously (see #324)

The failure is in `test_pyfftw_interfaces_cache.py::CacheTest::test_objects_removed_after_keepalive`:

<details>

```
=============================================================== FAILURES ================================================================
____________________________________________ CacheTest.test_objects_removed_after_keepalive _____________________________________________

self = <tests.test_pyfftw_interfaces_cache.CacheTest testMethod=test_objects_removed_after_keepalive>

    @unittest.skipIf(*miss('64'))
    def test_objects_removed_after_keepalive(self):
        _cache = interfaces.cache._Cache()
    
        key = 'the key'
    
        test_array = numpy.random.randn(16)
        obj = builders.fft(test_array)
        _cache.insert(obj, key)
    
        self.assertIs(_cache.lookup(key), obj)
    
        keepalive_time = _cache.keepalive_time
    
        if os.name == 'nt':
            # A hack to keep appveyor from falling over here. I suspect the
            # contention is too much to work properly. Either way, let's
            # assume it's a windows problem for now...
            time.sleep(keepalive_time * 8)
        else:
            # Relax a bit more otherwise
            time.sleep(keepalive_time * 4)
    
        self.assertRaises(KeyError, _cache.lookup, key)
    
        _cache.insert(obj, key)
        old_keepalive_time = _cache.keepalive_time
        _cache.set_keepalive_time(old_keepalive_time * 4)
    
        self.assertIs(_cache.lookup(key), obj)
    
        time.sleep(old_keepalive_time * 3.5)
        # still should be there
        self.assertIs(_cache.lookup(key), obj)
    
        if os.name == 'nt':
            # As above, but with a bit longer
            time.sleep(old_keepalive_time * 16)
        else:
            time.sleep(old_keepalive_time * 8)
    
>       self.assertRaises(KeyError, _cache.lookup, key)
E       AssertionError: KeyError not raised by lookup

test_pyfftw_interfaces_cache.py:328: AssertionError
=========================================================== warnings summary ============================================================
test_r2r.py:125
  /Users/glee/Downloads/pyFFTW_wheels/pyFFTW/tests/test_r2r.py:125: PytestCollectionWarning: cannot collect test class 'TestRealTransform' because it has a __init__ constructor (from: tests/test_r2r.py)
[...]

PR #336, opened 2021-12-22
Increase timing in cache tests

grlee77:
This PR increases sleep times during cache tests to try to avoid the (random) failures mentioned here:
https://github.com/pyFFTW/pyFFTW/issues/327#issuecomment-997200430


hgomersall (approved):

grlee77:
Okay, this has passed twice, so let's merge and see if we encounter any more failures going forward.

PR #338, opened 2021-12-23
add release notes for v0.13.0

grlee77:


hgomersall (approved):

hgomersall:
It feels to me this should be a 1.0rc. We can have it in the wild for a bit first though...

grlee77:
I am happy to update this to use 1.0 instead of 0.13 if you prefer.  I think we are covering all the transforms now and things are pretty stable, so that seems reasonable.

hgomersall:
Nah, let's stick with 0.13 for the moment and release when we're confident everything works properly. Thanks!

PR #345, opened 2022-03-02
fix typo

MicaelJarniac:
It was `> 3.7`, as in `x > 3.7`, which means that versions `x` greater than `3.7` are not supported.
I believe it was meant to be `< 3.7`, as in versions lower than `3.7` are not supported.


grlee77 (approved):
LGTM, thanks!

PR #359, opened 2023-07-17
Restrict Cython to <3 for now

rmorshea:
On M1 Macs, installing from source is broken. Likely due to the recent release of Cython 3.0. I've seen issues with Cython 3 being reported [more broadly](https://github.com/yaml/pyyaml/issues/601) though so I've restricted `build-system.requires` to `cython<3` for all platforms.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -167,7 +167,7 @@ test-command = "pytest --import-mode=append {project}"
 requires = [
     "wheel",
     "setuptools<=59.4.0",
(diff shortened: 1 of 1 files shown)


rmorshea:
PyYAML is making the exact same change: https://github.com/yaml/pyyaml/pull/702

paugier:
pyFFTW is now built with Cython 3 so this PR can be closed!

Issue #349, opened 2022-09-22
M1 installation issue


young-oct:
Issue description:  Failed to build pyfftw with pip install pyfftw command
System: MacBook Pro(M1) version 12.6
Description:  I encountered  "Failed to build pyfftw" when I first treid to install sporoco library, which was provided by @bwohlberg
I tried to install pyfftw manually with  pip install pyfftw but it still fails to build

`pip install pyfftw

Collecting pyfftw
  Using cached pyFFTW-0.13.0.tar.gz (113 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Requirement already satisfied: numpy<2.0,>=1.16 in /Users/youngwang/opt/anaconda3/envs/complex/lib/python3.10/site-packages (from pyfftw) (1.22.4)
Building wheels for collected packages: pyfftw
  Building wheel for pyfftw (pyproject.toml) ... error
  error: subprocess-exited-with-error
  
  × Building wheel for pyfftw (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [71 lines of output]
      running bdist_wheel
      running build
      running build_py
      creating build
      creating build/lib.macosx-11.0-arm64-3.10
      creating build/lib.macosx-11.0-arm64-3.10/pyfftw
      copying pyfftw/config.py -> build/lib.macosx-11.0-arm64-3.10/pyfftw
      copying pyfftw/_version.py -> build/lib.macosx-11.0-arm64-3.10/pyfftw
      copying pyfftw/__init__.py -> build/lib.macosx-11.0-arm64-3.10/pyfftw
      creating build/lib.macosx-11.0-arm64-3.10/pyfftw/builders
      copying pyfftw/builders/builders.py -> build/lib.macosx-11.0-arm64-3.10/pyfftw/builders
      copying pyfftw/builders/__init__.py -> build/lib.macosx-11.0-arm64-3.10/pyfftw/builders
      copying pyfftw/builders/_utils.py -> build/lib.macosx-11.0-arm64-3.10/pyfftw/builders
      creating build/lib.macosx-11.0-arm64-3.10/pyfftw/interfaces
      copying pyfftw/interfaces/cache.py -> build/lib.macosx-11.0-arm64-3.10/pyfftw/interfaces
      copying pyfftw/interfaces/__init__.py -> build/lib.macosx-11.0-arm64-3.10/pyfftw/interfaces
[...]

Issue #316, opened 2021-08-19
How to install pyFFTW in Mac M1?


MR-T77:
I'm installing a conda environment (with miniforge) in a MacOs M1 with tensorflow and other libraries that require pyFFTW, but I can't install pyFFTW. Any workaround?

> pip install pyfftw
Collecting pyfftw
  Using cached pyFFTW-0.12.0.tar.gz (137 kB)
Requirement already satisfied: numpy<2.0,>=1.10 in /Users/asapinto/miniforge3/envs/last/lib/python3.8/site-packages (from pyfftw) (1.21.2)
Building wheels for collected packages: pyfftw
  Building wheel for pyfftw (setup.py) ... error
  ERROR: Command errored out with exit status 1:
   command: /Users/asapinto/miniforge3/envs/last/bin/python3.8 -u -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/private/var/folders/0k/hz9yngm56nz1htdc3c3t3d0c0000gn/T/pip-install-t9_j9w8n/pyfftw/setup.py'"'"'; __file__='"'"'/private/var/folders/0k/hz9yngm56nz1htdc3c3t3d0c0000gn/T/pip-install-t9_j9w8n/pyfftw/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' bdist_wheel -d /private/var/folders/0k/hz9yngm56nz1htdc3c3t3d0c0000gn/T/pip-wheel-ls9ariz0
       cwd: /private/var/folders/0k/hz9yngm56nz1htdc3c3t3d0c0000gn/T/pip-install-t9_j9w8n/pyfftw/
  Complete output (25 lines):
  running bdist_wheel
  running build
  running build_py
  creating build
  creating build/lib.macosx-11.0-arm64-3.8
  creating build/lib.macosx-11.0-arm64-3.8/pyfftw
  copying pyfftw/config.py -> build/lib.macosx-11.0-arm64-3.8/pyfftw
  copying pyfftw/_version.py -> build/lib.macosx-11.0-arm64-3.8/pyfftw
  copying pyfftw/__init__.py -> build/lib.macosx-11.0-arm64-3.8/pyfftw
  creating build/lib.macosx-11.0-arm64-3.8/pyfftw/builders
[...]

PR #365, opened 2023-11-04
Add search paths for FFTW on macOS/Apple Silicon hosts

codyd51:
Hi! 

I tried to install pyFTTW on an M2 host and encountered some issues that I couldn't fix without modifying the project build. I see that some other people have run across this as well (such as in #349 and #316), so I thought I'd send the changes your way. I hope it's useful! Please let me know if there's anything here I should adjust. 

All the best.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -67,7 +67,7 @@ if os.environ.get("READTHEDOCS") == "True":
     os.environ["AR"] = "x86_64-linux-gnu-ar"
 
 
-def _get_mac_os_homebrew_prefix():
+def _get_mac_os_homebrew_prefix() -> str:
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -66,6 +67,25 @@
     os.environ["AR"] = "x86_64-linux-gnu-ar"
 
 
+def _get_mac_os_homebrew_prefix() -> str:
```
codyd51, line 70: My preference would be a `pathlib.Path` here. I didn't bother because it made the call sites below a bit wordy, and because it doesn't fit with the rest of the project style as far as I saw.

PR #369, opened 2024-01-17
Fix conflicting get_platform changes in setup.py

karlotness:
I think this should fix the build issues after combining #363 and #365.


karlotness:
Thanks!

PR #375, opened 2024-04-01
Proposition to allow execute in a nogil environment

### LINKED ISSUE #374: Executing Pyfftw in a nogil block
I think that should work fine assuming Cython is happy. I would suggest limiting yourself to the normal calls.

I think you can call `execute` with `nogil`. The innards of that are read-only properties so unless you're doing something nefarious to the class it should be ok. Is it more than just the execute part?

Have you noted that `execute` already runs as no-gil?

_Originally posted by @hgomersall in https://github.com/pyFFTW/pyFFTW/issues/373#issuecomment-1966291886_

gabrielfougeron:
Closes #374

This is my proposition to allow running a fft from the inside of a nogil environment.

What I did :

- Create two new attributes pointing to the start of  `_input_array` and `_output_array`
```cython
    cdef void* _input_pointer
    cdef void* _output_pointer
```

- Make sure that `_input_pointer` and `_output_pointer` always do point to the start of  `_input_array` and `_output_array`.

- Broke down `execute` in two with a new `cdef void execute_nogil(self) noexcept nogil`

- Declare everything that is cdef/cpdef in `cdef class FFTW` in `pyfftw.pxd` in order to be able to cimport the cdef class from outside pyfftw. 
- Add a simple function `execute_in_nogil` for test purposes
- Add a simple test in file `test_pyfftw_nogil`

Don't hesitate to tell me what you think


pyfftw/pyfftw.pxd
```diff
@@ -403,3 +403,65 @@ cdef enum:
     FFTW_PATIENT = 32
     FFTW_ESTIMATE = 64
     FFTW_WISDOM_ONLY = 2097152
+
```
hgomersall, line 406: Do we need all these attributes to be exposed?

pyfftw/pyfftw.pyx
```diff
@@ -1993,15 +2006,23 @@ cdef class FFTW:
...
+            self.execute_nogil()
+
+    cdef void execute_nogil(self) noexcept nogil:
+
+        self._fftw_execute(
+            self._plan,
+            self._input_pointer,
+            self._output_pointer)
+
+cpdef void execute_in_nogil(FFTW FFTW_object):
```
hgomersall, line 2019: I don't fully understand the need for this function, nor for the associated test...

pyfftw/pyfftw.pxd
gabrielfougeron: Previously, the cdef class FFTW was not declared in a *pxd file. This meant I could not type cimport the associated type from outside the library. With this declaration, now I can.

As for the attributes themselves, as far as I can tell (from Cython complaining), you need all of them to be present in the declaration, as well as all cdef / cpdef methods. def methods don't have to be there.

pyfftw/pyfftw.pyx
gabrielfougeron: Good point. 
I went back and forth a few times on how best to implement the cdef class method. I honestly though initially that those could only be accessed from within the class (private methods). This function is there merely to demonstrate that this is not the case. It adds no functionality and can be removed.

gabrielfougeron:
Could you please run the test suite and make sure that everything goes well?
I ran the tests on my machine before and after my additions, and I didn't see much difference, but still some tests failed and some were skipped, and I don't have the codebase knowledge to certify that everything is OK.

tests/test_pyfftw_nogil.py
```diff
@@ -0,0 +1,105 @@
+# Copyright 2014 Knowledge Economy Developments Ltd
+#
+# Henry Gomersall
```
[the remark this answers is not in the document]
gabrielfougeron: You are right, better safe than sorry.
This point is moot anyway since I just removed the file.

gabrielfougeron:
Hi Henry,

Before my new push, I could do the following in my code:

```cython
cimport pyfftw

cpdef void object_execute_in_nogil(pyfftw.FFTW fftw_object):

    with nogil:
        fftw_object.execute_nogil()
```

But what I **really** wanted was to run several different FFTs. This is a problem because I cannot put several FFTW objects in a list and access items of the list in a nogil environment.

I made a few additional modifications to the PR:

- I defined and exposed in the Cython C interface a new ctypedef struct containing an execute function pointer, a plan and pointers to input and output arrays.
- I defined a getter in your cdef class FFTW for that struct and a simple function to run the fft. This allows me to write the following in my code:

```cython
cimport pyfftw
from libc.stdlib cimport malloc, free

cdef class mystruct:

    cdef pyfftw.fftw_exe** exe_arr

    def pyfftw_list_to_carr(self, pyfftw_list):

        cdef Py_ssize_t nfft = len(pyfftw_list)
        cdef Py_ssize_t i

        cdef pyfftw.FFTW fftw_object
        self.exe_arr = <pyfftw.fftw_exe**> malloc(sizeof(pyfftw.fftw_exe*) * nfft)

        for i in range(nfft):
            fftw_object = pyfftw_list[i]
            self.exe_arr[i] =  <pyfftw.fftw_exe*> malloc(sizeof(pyfftw.fftw_exe))
            self.exe_arr[i][0] = fftw_object.get_fftw_exe()

```

- Untelated : I included the contents of `pyfftw_complex.h` directly in  `pyfftw_complex.h` because Cython would complain about not knowing where the file is when linking another project with pyfftw.

hgomersall:
> ```cython
> ```cython
> 
> ```
> 
> 
>     
>       
>     
> 
>       
>     
> 
>     
>   
> But what I really wanted was to run several different FFTs. This is a problem because I cannot put several FFTW objects in a list and access items of the list in a nogil environment.
> ```

Is there not a lower level data structure that can be used to store the FFTW objects?

I'm wary of the approach you've proposed as it allows multiple non-locked instances of the underlying data. It would be easy for different calls to `execute_in_nogil` to clobber one another. In fact, that's the case anyway for accessing self.

I think a better solution would be to wrap the inner data structure in a mutex. It should be very cheap if there's no conflict, but it would ensure data integrity. Essentially, the nogil route should be exclusive for a given array.

gabrielfougeron:
> I think a better solution would be to wrap the inner data structure in a mutex. It should be very cheap if there's no conflict, but it would ensure data integrity. Essentially, the nogil route should be exclusive for a given array.

I think I understand what your safety concern is, although I do not know what a mutex is ....

hgomersall:
> > I think a better solution would be to wrap the inner data structure in a mutex. It should be very cheap if there's no conflict, but it would ensure data integrity. Essentially, the nogil route should be exclusive for a given array.
> 
> I think I understand what your safety concern is, although I do not know what a mutex is ....

A mutual exclusion - essentially a way for an aliased data structure to be accessed exclusively - the mutex can only be claimed in a single context, so multiple attempts to lock it will block until the previous one is released.

gabrielfougeron:
I've never done this, so before I have a go at implementing a mutex lock to prevent concurrent calls to this low-level execute, let me please try to clarify a few points. Please do contradict me when I'm wrong.

- The problem with the current implementation is that several threads could use the newly defined `execute_in_nogil` concurrently leading to absurd results. Preventing this is the work of the GIL in Python, but since the whole point of this PR is to work around the GIL, a mutex seems like the way to go.
- In C, mutex are implemented in the pthread library.
- I see mentions of pthread in your setup.py. cf this part of the code:
https://github.com/pyFFTW/pyFFTW/blob/82ae9eafac5fdd411f38852a1d379bb013526460/setup.py#L273-L281
I seems to me like the C pthread library is only a dependency if FFTW is detected to have been built with pthread. Is this correct?
 - If so, should pthread now be a dependency regardless, or should I check for the env variable 'PYFFTW_USE_PTHREADS' to lock the data?
 - What about OMP-enabled FFTW ?

gabrielfougeron:
Just FYI, with the latest push, I managed to include pyfftw as an optionnal dependency of my own project https://github.com/gabrielfougeron/choreo . All tests are passing and the results are consistent with mkl-fft and pocketfft.

hgomersall:
I think probably this is fine now. It would be good if you could add a little bit of documentation to the nogil method to make it clear that it is not threadsafe and needs locking. Once we're happy with the tests passing we can merge.

gabrielfougeron:
Done :-)

gabrielfougeron:
Fixed bug: External package using Cython headers would only build if pyfftw was installed in editable mode.
Including cython pxd headers in distributed wheels fixed the problem.

hgomersall:
@paugier would you mind quickly reviewing this in light of the recent changes? Particularly the changes to `setup.py` etc. I think broadly the implementation is fine.

paugier:
Error on Windows: 

```
2024-07-08T11:15:59.2728785Z   error in setup command: "values of 'package_data' dict" must be a list of strings (got ['*.h', '*.pxd', ['libfftw3-3.dll', 'libfftw3l-3.dll', 'libfftw3f-3.dll']])
```

which seems related to a change in setup.py.

setup.py
```diff
@@ -122,8 +123,8 @@ def get_package_data():
             pass
         else:
             # as download from http://www.fftw.org/install/windows.html
-            package_data['pyfftw'] = [
-                'libfftw3-3.dll', 'libfftw3l-3.dll', 'libfftw3f-3.dll']
+            package_data['pyfftw'].append([
```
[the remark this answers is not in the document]
gabrielfougeron: Indeed, silly mistake!

paugier (approved):

hgomersall (approved):

Issue #346, opened 2022-04-20
Segfault in libfftw3f-19f177d3.so.3.2.3[xxx+ef000]


phlibi:
Hello everyone,

first off, this may likely not be a bug in PyFFTW, but something wrong with my software. I suppose this is the best way to reach one of the experts on the topic.

Short description of the project: We acquire holograms using machine-vision cameras plus many other event-based information. All these event objects are independent and dispatched to a number of threads, which perform various calculations. Among others, the holograms are reconstructed at various depths using FFT operations. This is where PyFFTW comes into play. Since there are many other operations performed, we do the parallelization in our software and run PyFFTW with only a single thread. Same holds for the caching.

All this has been deployed to several systems and works very well most of the time. However, especially under heavy load (10s of events per second), there are sporadic segmentation faults of the Python container (Docker), all pointing to the same spot in libfftw3f:
```
$ dmesg -H | grep seg
[Apr17 21:12] python3[1388111]: segfault at 1c ip 00007f8976331b2c sp 00007f88967fa5f8 error 4 in libfftw3f-19f177d3.so.3.2.3[7f89762cb000+ef000]
[Apr18 16:26] python3[1764720]: segfault at 1c ip 00007f68f8afeb2c sp 00007f68057f85f8 error 4 in libfftw3f-19f177d3.so.3.2.3[7f68f8a98000+ef000]
[Apr18 16:44] python3[2073605]: segfault at 0 ip 00007faa04bccb9c sp 00007fa9357f8630 error 4 in libfftw3f-19f177d3.so.3.2.3[7faa04ba1000+ef000]
[Apr18 17:58] python3[2078442]: segfault at 0 ip 00007ffab84e2b9c sp 00007ff9f57f8630 error 4 in libfftw3f-19f177d3.so.3.2.3[7ffab84b7000+ef000]
[Apr18 19:32] python3[2098405]: segfault at 0 ip 00007f7ea6c11c28 sp 00007f7e08410138 error 4 in libfftw3f-19f177d3.so.3.2.3[7f7ea6bce000+ef000]
[Apr20 05:35] python3[2492761]: segfault at 6 ip 00007f49398aeb9c sp 00007f48657f8630 error 4 in libfftw3f-19f177d3.so.3.2.3[7f4939883000+ef000]
```
I would be very surprised if anyone could simply diagnose the problem without further investigation, but I hope to get some hints about where to start.

- How can I find out in which function of FFTW the segfault happens?
[...]

PR #376, opened 2024-05-08
Possible fix of issue #346

AlexeyBarsuk:


hgomersall:
This seems like a good and simple solution. Does it solve your requirements?

AlexeyBarsuk:
It looks it does. I was not able to reproduce crush (#346) and I get expected results when using cache with multiple threads ( #376)

hgomersall:
Hmmm, tests are failing...

AlexeyBarsuk:
Can't imagine how changes from my commits can break tests.  Commit with lock on plan destruction should not influence any way to results of FFT routines. Moreover the lock should be there.
As about the second commit - there are assertions in tests that do not use cache at all. For example 
```
   /project/tests/test_pyfftw_scipy_fft.py:194: AssertionError
  ____________ InterfacesScipyR2RFFTNTestDSTNFloat64.test_normalized _____________
```
 So this commit does not influence to these tests.
 
From the other side I see that scipy version installed for tests is the newest. 
```
 Collecting scipy
    Downloading scipy-1.13.0-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl.metadata (60 kB)
```
Maybe it is too brave assumption, but in case of non fixed versions of external libraries in tests, the problems with tests may be related to new versions of external libraries. Any way I do not have any reasonable assumption why tests are failed.

hgomersall:
I don't think its your changes, I just think the tests are broken, which is a problem in its own right.

paugier:
I think this just needs to be rebased. For simplicity, I'm going to do it in a new PR.

hgomersall:
Changes merged into #381

PR #370, opened 2024-01-25
Upgrade setuptools and configparser for Python 3.12

jsalort:
The current version of PyFFTW does not install with Python 3.12, see [Issue 336](https://github.com/pyFFTW/pyFFTW/issues/366).
This PR proposes to lift the setuptools version constraint, and upgrade versioneer to replace SafeConfigParser which was removed in Python 3.12.
I had also to add "py_modules" explicitely in `setup.py`, and change a string to a raw-string in versioneer because of a non-escaped '\s' in the string.

PyFFTW now builds and installs fine, but I get a number of pytest failures,
```
===================================================================== short test summary info =====================================================================
FAILED tests/test_pyfftw_builders.py::BuildersTestFFT::test_auto_align_input - AssertionError: False is not true
FAILED tests/test_pyfftw_builders.py::BuildersTestIFFT::test_auto_align_input - AssertionError: False is not true
FAILED tests/test_pyfftw_builders.py::BuildersTestRFFT::test_auto_align_input - AssertionError: False is not true
FAILED tests/test_pyfftw_builders.py::BuildersTestIRFFT::test_auto_align_input - AssertionError: False is not true
FAILED tests/test_pyfftw_builders.py::BuildersTestFFT2::test_auto_align_input - AssertionError: False is not true
FAILED tests/test_pyfftw_builders.py::BuildersTestIFFT2::test_auto_align_input - AssertionError: False is not true
FAILED tests/test_pyfftw_builders.py::BuildersTestRFFT2::test_auto_align_input - AssertionError: False is not true
FAILED tests/test_pyfftw_builders.py::BuildersTestIRFFT2::test_auto_align_input - AssertionError: False is not true
FAILED tests/test_pyfftw_builders.py::BuildersTestFFTN::test_auto_align_input - AssertionError: False is not true
FAILED tests/test_pyfftw_builders.py::BuildersTestIFFTN::test_auto_align_input - AssertionError: False is not true
FAILED tests/test_pyfftw_builders.py::BuildersTestRFFTN::test_auto_align_input - AssertionError: False is not true
FAILED tests/test_pyfftw_builders.py::BuildersTestIRFFTN::test_auto_align_input - AssertionError: False is not true
FAILED tests/test_pyfftw_class_misc.py::FFTWMiscTest::test_aligned_flag - AssertionError: False is not true
FAILED tests/test_pyfftw_complex.py::Complex128FFTWTest::test_alignment - AssertionError: False is not true
[...]

PR #380, opened 2024-06-27
Support Python 3.12

paugier:
This is basically #370 rebased on master with changes to build wheels for Py 3.12 on Github Actions.

### DIFF
--- a/.github/workflows/wheel_tests_and_release.yml
+++ b/.github/workflows/wheel_tests_and_release.yml
@@ -21,7 +21,7 @@ jobs:
       matrix:
         os: [ubuntu-22.04]
 
-        cibw_python: ["cp39", "cp310", "cp311", "pp39"]
+        cibw_python: ["cp39", "cp310", "cp311", "cp312", "pp39"]
         # SciPy and NumPy don't support musllinux
         cibw_libc: ["manylinux"]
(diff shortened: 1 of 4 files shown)


hgomersall:
These failures look rather odd...

paugier:
The builds are failing only on win_arm64 with

```
File "C:\Users\runneradmin\AppData\Local\Temp\pip-build-env-3ccuw6go\overlay\Lib\site-packages\setuptools\_distutils\_msvccompiler.py", line 241, in initialize
        assert not self.initialized, "don't init multiple times"
```

I noted that `numpy-1.19.5-cp39-cp39-win_amd64.whl` is used, which does not make sense.

paugier:
It seems that at least numpy, scipy, scikit-learn and scikit-image do not build win_amd64 wheels. It might be reasonable to do the same? What do you think @hgomersall ?

paugier:
> Is this only on python 3.12?

No it's for all supported Python version. I guess it's an issue with the more recent setuptools that is needed for Python 3.12.

I remove win_arm64 from the matrix to follow numpy and scipy. It can be added latter when the community better supports arm64 on Windows.

paugier:
@hgomersall, I thing this PR is ready to be reviewed!

hgomersall (approved):

PR #390, opened 2024-07-09
Add docs/release/0.14.0.rst

paugier:
Needed for #385

PR #398, opened 2024-10-22
Query the version and compiler flags of fftw

### LINKED ISSUE #355: Feature request: Export fftw_version
When pyFFTW is installed, it is fairly hard to tell the version of the underlying FFTW library matching their [release notes](https://fftw.org/release-notes.html). The version code in the shared library file name (e.g. libfftw3f.so.3.6.10 for FFTW 3.3.10) is *not* the release version, but follows the [Libtool versioning scheme](https://www.gnu.org/software/libtool/manual/html_node/Libtool-versioning.html). In request to this, Matteo Frigo told me that FFTW actually exports its version as `fftw_version` variable:

> printf("%s\n", fftw_version)  will tell you which version you are using.  See also fftw_cc for which compiler was used and which flags.

Feature request: Make `fftw_version` accessible in the Python module (and maybe also `fftw_cc`).

ptim0626:
Fixes #355.

This is my first PR to pyfftw,  please let me know if I have missed anything, and I also want to take this chance to thank all the developers in pyfftw!

This PR exposes `fftw*_version`, `fftw*_cc` and `fftw*_codelet_optim` of the fftw library, addressing #355.

If users are interested in the version of fftw library, this can be queried in the high-level API, e.g.
```python
import pyfftw
print(pyfftw.fftw_version)
```
For easier comparison, the fftw version is also available as a 3-member tuple as `(major, minor, release)` (idea borrowed from h5py), e.g.
```python
import pyfftw
if pyfftw.fftw_version_tuple > (3, 3, 5):
    # do something if the fftw library version is newer than 3.3.5 
    ...
```
How the fftw library was compiled can be inspected by
```python
import pyfftw
print(pyfftw.fftw_cc)
```
For some rare cases, users may be interested in information from different precisions (float, double, long double), this is also available as dictionaries:
```python
import pyfftw
print(pyfftw.pyfftw._fftw_version_dict)
print(pyfftw.pyfftw._fftw_cc_dict)
print(pyfftw.pyfftw._fftw_codelet_optim_dict) # options for compiling the codelet
```
Ready for review.

### DIFF
--- a/pyfftw/pyfftw.pyx
+++ b/pyfftw/pyfftw.pyx
@@ -69,19 +69,23 @@ if np.dtype(np.longdouble) != np.dtype(np.float64):
 _supported_types = []
 _supported_nptypes_complex = []
 _supported_nptypes_real = []
+_fftw_version_dict = {}
 
 if PYFFTW_HAVE_SINGLE:
     _supported_types.append('32')
     _supported_nptypes_complex.append(np.complex64)
     _supported_nptypes_real.append(np.float32)
+    _fftw_version_dict['32'] = fftwf_version.decode()
 if PYFFTW_HAVE_DOUBLE:
     _supported_types.append('64')
     _supported_nptypes_complex.append(np.complex128)
     _supported_nptypes_real.append(np.float64)
+    _fftw_version_dict['64'] = fftw_version.decode()
 if PYFFTW_HAVE_LONG:
     _supported_types.append('ld')
     _supported_nptypes_complex.append(np.clongdouble)
     _supported_nptypes_real.append(np.longdouble)
+    _fftw_version_dict['ld'] = fftwl_version.decode()
 
 if (PYFFTW_HAVE_SINGLE_OMP or PYFFTW_HAVE_DOUBLE_OMP or PYFFTW_HAVE_LONG_OMP):
     _threading_type = 'OMP'
--- a/pyfftw/pyfftw.pyx
+++ b/pyfftw/pyfftw.pyx
@@ -45,11 +45,7 @@ import platform
 import warnings
 import threading
 
-# the fftw .dll (3.3.5) does not export fftw*version, fftw*_cc and
-# fftw*_codelet_optim etc., so for Windows the version is hard-coded to
-# 3.3.5 and others as 'NOT_AVAILABLE'
 _ON_WINDOWS = platform.system() == 'Windows'
-_WINDOWS_FFTW_VERSION = '3.3.5'
 
 include 'utils.pxi'
 
@@ -80,12 +76,15 @@ _fftw_version_dict = {}
 _fftw_cc_dict = {}
 _fftw_codelet_optim_dict = {}
 
+# the fftw .dll (as of 3.3.5) does not export fftw*version, fftw*_cc and
+# fftw*_codelet_optim etc., so for Windows those information are simply
+# stated as 'NOT_AVAILABLE'
 if PYFFTW_HAVE_SINGLE:
     _supported_types.append('32')
     _supported_nptypes_complex.append(np.complex64)
     _supported_nptypes_real.append(np.float32)
     if _ON_WINDOWS:
-        _fftw_version_dict['32'] = _WINDOWS_FFTW_VERSION
+        _fftw_version_dict['32'] = 'NOT AVAILABLE'
         _fftw_cc_dict['32'] = 'NOT AVAILABLE'
         _fftw_codelet_optim_dict['32'] = 'NOT AVAILABLE'
     else:
@@ -97,7 +96,7 @@ if PYFFTW_HAVE_DOUBLE:
     _supported_nptypes_complex.append(np.complex128)
     _supported_nptypes_real.append(np.float64)
     if _ON_WINDOWS:
-        _fftw_version_dict['64'] = _WINDOWS_FFTW_VERSION
+        _fftw_version_dict['64'] = 'NOT AVAILABLE'
         _fftw_cc_dict['64'] = 'NOT AVAILABLE'
         _fftw_codelet_optim_dict['64'] = 'NOT AVAILABLE'
     else:
@@ -109,7 +108,7 @@ if PYFFTW_HAVE_LONG:
     _supported_nptypes_complex.append(np.clongdouble)
     _supported_nptypes_real.append(np.longdouble)
     if _ON_WINDOWS:
-        _fftw_version_dict['ld'] = _WINDOWS_FFTW_VERSION
+        _fftw_version_dict['ld'] = 'NOT AVAILABLE'
         _fftw_cc_dict['ld'] = 'NOT AVAILABLE'
         _fftw_codelet_optim_dict['ld'] = 'NOT AVAILABLE'
     else:
--- a/pyfftw/pyfftw.pyx
+++ b/pyfftw/pyfftw.pyx
@@ -1988,10 +1988,10 @@ cdef class FFTW:
 
         **For Cython use only.**
 
-        This is really only useful if you want to 
+        This is really only useful if you want to
         bundle a few :data:`pyfftw.fftw_exe` in a C array, and then call them all from
         within a nogil block.
-        
+
         '''
 
         cdef fftw_exe exe
@@ -2011,7 +2011,7 @@ cdef void execute_in_nogil(fftw_exe* exe_ptr) noexcept nogil:
     **For Cython use only.**
 
     Warning: This method is **NOT** thread-safe. Concurrent calls
-    to :func:`pyfftw.execute_in_nogil` with an aliased :data:`pyfftw.fftw_exe` will lead 
+    to :func:`pyfftw.execute_in_nogil` with an aliased :data:`pyfftw.fftw_exe` will lead
     to wrong FFT results.
 
     '''
--- a/pyfftw/pyfftw.pyx
+++ b/pyfftw/pyfftw.pyx
@@ -70,22 +70,26 @@ _supported_types = []
 _supported_nptypes_complex = []
 _supported_nptypes_real = []
 _fftw_version_dict = {}
+_fftw_cc_dict = {}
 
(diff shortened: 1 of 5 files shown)


hgomersall:
I think it looks good to me. @paugier happy with it? Presumably we should finish the release prior to this?

hgomersall:
@ptim0626 Thanks for this!

paugier:
Nice addition. Note that the tuple can be used  as

```
if pyfftw.fftw_version_tuple > (3, 2, 0):
    ...
```

Maybe such test can be added to the test. It's a bit more functional than just testing the length.

Also I guess it would make sense to merge this PR after https://github.com/pyFFTW/pyFFTW/pull/390.

tests/test_pyfftw_version.py
```diff
@@ -0,0 +1,28 @@
...
+    else:
+        # cannot get version from fftw
+        assert True
+
+
+def test_fftw_version_tuple():
+    fvt = pyfftw.fftw_version_tuple
+    if fvt:
+        # major, minor, release
+        assert len(fvt) == 3
```
paugier, line 20: I what thinking about something like that:

```suggestion
        assert fvt > (0, 0, 0)
```

paugier:
There is an error for the CI on Windows:

```
2024-10-23T12:15:12.4519763Z      Creating library build\temp.win-amd64-cpython-311\Release\pyfftw\pyfftw.cp311-win_amd64.lib and object build\temp.win-amd64-cpython-311\Release\pyfftw\pyfftw.cp311-win_amd64.exp
2024-10-23T12:15:12.4776370Z   pyfftw.obj : error LNK2001: unresolved external symbol __imp_fftw_version
2024-10-23T12:15:12.4779454Z   pyfftw.obj : error LNK2001: unresolved external symbol __imp_fftwf_codelet_optim
2024-10-23T12:15:12.4782621Z   pyfftw.obj : error LNK2001: unresolved external symbol __imp_fftw_codelet_optim
2024-10-23T12:15:12.4784556Z   pyfftw.obj : error LNK2001: unresolved external symbol __imp_fftwl_cc
2024-10-23T12:15:12.4786486Z   pyfftw.obj : error LNK2001: unresolved external symbol __imp_fftwf_version
2024-10-23T12:15:12.4788342Z   pyfftw.obj : error LNK2001: unresolved external symbol __imp_fftwf_cc
2024-10-23T12:15:12.4790227Z   pyfftw.obj : error LNK2001: unresolved external symbol __imp_fftwl_version
2024-10-23T12:15:12.4792028Z   pyfftw.obj : error LNK2001: unresolved external symbol __imp_fftw_cc
2024-10-23T12:15:12.4793910Z   pyfftw.obj : error LNK2001: unresolved external symbol __imp_fftwl_codelet_optim
2024-10-23T12:15:12.4796356Z   build\lib.win-amd64-cpython-311\pyfftw\pyfftw.cp311-win_amd64.pyd : fatal error LNK1120: 9 unresolved externals
2024-10-23T12:15:12.5240080Z   error: command 'C:\\Program Files\\Microsoft Visual Studio\\2022\\Enterprise\\VC\\Tools\\MSVC\\14.41.34120\\bin\\HostX86\\x64\\link.exe' failed with exit code 1120
```

paugier:
@ptim0626  can you please rebase on master so that the CI is rerun with the versions officially supported (3.10 to 3.13) ?

Of course, we also need to fix the issue on Windows. @ptim0626 do you have an idea of what happens and what should be done ?

ptim0626:
Hi @paugier, sorry I was distracted by other stuff. I will address your comments and rebase my work on the new master to check.

About the Windows error, the symbols `fftw_version`, `fftw_cc` and `fftw_codelet_optim` etc are not exported correctly. From the `fftw3.h` header file, the relevant part is here

https://github.com/pyFFTW/pyFFTW/blob/06830e1aefc2a31791e88244b547eaf8dda61386/include/win/fftw3.h#L348-L350

For Linux, `FFTW_EXTERN` is just the ordinary `extern`, but for Windows it is something else and I am afraid I am not familiar with how this should be done, but I will investigate whether I export the symbol correctly in the first place, as other functions are exported correctly (in Windows).

tests/test_pyfftw_version.py
ptim0626: @paugier Thanks for the suggestion, I agree the test would be more helpful by checking the content of the tuple. I think if pyfftw supports only fftw 3, would `fvt > (3, 0, 0)` be a good choice?

Thanks for pointing out that tuple can be compared in this way!

ptim0626:
Looking at the CI workflow it seems it downloads the fftw library and it may not have the macro `FFTW_DLL` defined, which is required for importing/exporting symbols in Windows correctly.

The header `pyFFTW/include/win/fftw3.h` has

https://github.com/pyFFTW/pyFFTW/blob/51dccc55755d7bbac8e63a5d139390a3fd73ffbd/include/win/fftw3.h#L71-L89

But the downloaded fftw library may not. I modified the `setup.py` to define this macro and hopefully this will work.

Btw does the `pyFFTW/include/win/fftw3.h` header file being used?

ptim0626:
Unfortunately my change (defining the macro `FFTW_DLL`) seems useless, it gives the same error. I don't have a Windows machine to play with it myself, perhaps I will try setting this up with some containers and understand the issue.

Meanwhile if anyone has any hint about what is happening I would be very grateful!

ptim0626:
I guess the problem may be related to the DLL files for Windows. When I looked at the `.def` and `.dll` of https://fftw.org/pub/fftw/fftw-3.3.5-dll64.zip (used in the CI), it seems it didn't export `fftw_version`, `fftw_cc` and `fftw_codelet_optim` etc.

For example, the symbol `fftw_version` is defined in `libfftw3-3.dll`,
```console
$ nm --demangle libfftw3-3.dll | grep 'version'
000000007085df80 R fftw_version
0000000000000000 A __major_image_version__
0000000000000004 A __major_os_version__
0000000000000005 A __major_subsystem_version__
0000000000000000 A __minor_image_version__
0000000000000000 A __minor_os_version__
0000000000000002 A __minor_subsystem_version__
```
But the `libfftw3-3.def` file doesn't export it,
```console
$ grep 'version' libfftw3-3.def
# nothing
```
And similar for other symbol I added.

I am not sure by just by adding `fftw_version` under the `EXPORTS` section of `libfftw3-3.def` will make it works, without rebuilding the DLL files.

hgomersall:
Can we put this behind a preprocessor directive to run different paths based on architecture, and redefine internally as appropriate?

ptim0626:
@hgomersall Yes I agree, and it looks like the latest fftw for Windows is 3.3.5 and support is unofficial. If we build pyfftw with fftw 3.3.5, we can hard-code this. It is just the compiler flag `fftw*_cc` and `fftw*_codelet_optim` won't be available.

hgomersall:
Would you mind spinning up something like what you think would work?

ptim0626:
@hgomersall I updated the code handle situation in Windows which I think it will work, but I can't really test it so I will rely on the CI to tell me if it is working. If it works, I can add some tests afterwards.

Although fftw doesn't support Windows, I think I will let them know those symbols are not exported in Windows.

pyfftw/pyfftw.pyx
```diff
@@ -41,9 +41,16 @@ from libc.stdlib cimport calloc, malloc, free
...
 
+import platform
 import warnings
 import threading
 
+# the fftw .dll (3.3.5) does not export fftw*version, fftw*_cc and
+# fftw*_codelet_optim etc., so for Windows the version is hard-coded to
+# 3.3.5 and others as 'NOT_AVAILABLE'
+_ON_WINDOWS = platform.system() == 'Windows'
+_WINDOWS_FFTW_VERSION = '3.3.5'
```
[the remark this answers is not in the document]
ptim0626: @paugier 3.3.5 is the latest FFTW Windows DLL provided via fftw.org so I just chose 3.3.5. Of course people can build their own DLL with different version of fftw, so perhaps it is better to list it as `NOT_AVAILABLE` altogether as this reflects the fact that it is not available from the underlying fftw library directly. What do you think?

setup.py
```diff
@@ -650,6 +650,8 @@ def build_extensions(self):
         if get_platform() in ('win32', 'win-amd64'):
             # because Numpy 2.0 includes complex.h
             self._pyfftw_define_macros.append(("FFTW_NO_Complex", "1"))
+            # for import/export symbols correctly
```
paugier, line 653: Is it really useful? Anyway, we cannot get version data on Windows?

pyfftw/pyfftw.pyx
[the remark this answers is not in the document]
paugier: I would prefer not to give a potentially wrong information, so I think `"NOT_AVAILABLE"` make sense when the data is not available.

This lack of data on Windows is annoying. It means that  code like `if pyfftw.fftw_version_tuple > (3, 0, 0):` cannot be used in practice because it could be wrong on Windows.

However, is there a practical reason to use such checks?

paugier:
I was not able to measure the effect on import time so (as expected) it is completely negligible!

paugier:
The conflict in setup.py needs to be resolved, but it's very easy.

setup.py
ptim0626: Apology my bad, this was something I thought it might be helpful on building wheel in Windows. It turns out this is only relevant when building the .dll. Reverted.

pyfftw/pyfftw.pyx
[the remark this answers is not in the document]
ptim0626: @paugier I agree this is a bit annoying. I have emailed the authors of fftw about this, as I think this is an issue upstream (I may be wrong). Unless they provide a fftw .dll with those symbols exported, we can't reliably query the fftw version.

In some cases checking the fftw version may be helpful if the user depends on new features or bug fixes that are only available on some later versions. The compiler flag may be also useful in knowing whether fftw is complied with certain features.

paugier (approved):

PR #407, opened 2024-11-11
No need for setuptools at runtime

paugier:
I noticed that setuptools is mentioned as a runtime dependency and I don't understand why it would be needed. Let's try without.


paugier:
I think this is ready and that this small change could be merged.

hgomersall (approved):

PR #411, opened 2025-08-22
Fix CI

### LINKED ISSUE #313: setup.py does not use environment variable `LDFLAGS`.

### LINKED ISSUE #413: CI is broken (segmentation fault on macOS)
Currently, the CI is broken. 

First, macos-12 is no longer supported bu GA. Then, we need https://github.com/pyFFTW/pyFFTW/pull/411/commits/fcbbab3ffb65f48aa402145f472e59342179b700

After https://github.com/pyFFTW/pyFFTW/pull/411, there is still an issue on macOS x86_64 (3.11 and 3.12, but not 3.13!).

I don't know if it is related to the argument `-fopenmp` added from `setup.py` giving the error `clang: error: unsupported option '-fopenmp'`. It might be related but this error is also present for all macOS builds.

### LINKED ISSUE #414: SetuptoolsDeprecationWarning: License classifiers are deprecated.
```
SetuptoolsDeprecationWarning: License classifiers are deprecated.
  !!
  
          ********************************************************************************
          Please consider removing the following classifiers in favor of a SPDX license expression:
  
          License :: OSI Approved :: BSD License
  
          See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
          ********************************************************************************
  
  !!
```

paugier:
Fix #413
Fix #313
Fix #414

- Use macos-13 instead of macos-12 since it has been stopped by GitHub Actions.
- Stop support for Python 3.10 (following Numpy 2.3.2). (3.14 will be added in another PR.)
- Fix OpenMP builds for macOS (no more "error: -fopenmp not supported").
- Disable test for macOS x86_64 Python 3.11 and 3.12 due to mysterious segmentation faults :slightly_frowning_face: (associated issue: https://github.com/pyFFTW/pyFFTW/issues/418)
- Fix Nox (used to build the documentation).
- Update PDM lock file.
- [pdm: use pylock.toml (PEP 751)](https://github.com/pyFFTW/pyFFTW/pull/411/commits/e062d09460dcb099bc18c266c336053df1127485)


paugier:
I think this PR is ready. It would be nice to have a review since without a green CI, it's harder to work on other issues (see https://github.com/orgs/pyFFTW/projects/2).

hgomersall:
Sorry, been away.

I think it should still work on 3.10 - does this PR have a hard restriction on that? It looks good, but it would be useful I think to _allow_ 3.10.

paugier:
Shouldn't we follow https://scientific-python.org/specs/spec-0000/, as most core packages of the scientific Python ecosystem?

If we follow https://scientific-python.org/specs/spec-0000, the next pyfftw version should drop support for Python 3.10.

I don't really see the advantage of supporting Python 3.10 for the next version of pyfftw and that it is better to follow the ecosystem.

hgomersall:
Okeydoke. Let's stick with 0.16, and do a 1.0 release from a working release. I'm happy with this I think.

hgomersall (approved):

Issue #409, opened 2025-07-30
Add support for numpy 2.0 `out` argument in pyfftw.interfaces.numpy


maurerv:
## Summary

This issue proposes adding support for the `out` argument introduced in numpy 2.0 to the `pyfftw.interfaces.numpy` module. This would allow users to reuse pre-allocated buffers, saving computation time in scenarios involving sequential FFT computations.

## Motivation

I have a use case requiring sequential computation of large Fourier transforms where array buffers can be reused across iterations. While builders already support `out` arguments, managing multiple built objects of different shapes and parameters becomes cumbersome and hard to follow. 

For my case, the pyFFTW cache with the numpy interface yields comparable performance to builders and noticeably reduces code complexity. The only downside is the overhead associated with allocating new array objects to write forward and reverse Fourier transforms to, which could be avoided by supporting the `out` argument.

## Proposed Implementation

The initial implementation requires modifications to two locations in `pyfftw.interfaces._utils.Xfftn` and the corresponding interface functions:

1. Add `output_array` parameter to the `_Xfftn` function signature ([line 66](https://github.com/maurerv/pyFFTW/blob/59acacb60786785fab157f72e53916c57e65550d/pyfftw/interfaces/_utils.py#L66))

2. Implement logic to use provided output buffer instead of allocating new arrays ([line 189](https://github.com/maurerv/pyFFTW/blob/59acacb60786785fab157f72e53916c57e65550d/pyfftw/interfaces/_utils.py#L189))

3. Update interface functions to accept and pass through the `out` parameter ([example](https://github.com/maurerv/pyFFTW/blob/59acacb60786785fab157f72e53916c57e65550d/pyfftw/interfaces/numpy_fft.py#L275))

## Testing Strategy

I extended `InterfacesNumpyFFTTestFFT` to include tests for the proposed `out` argument:

1. Add an optional out parameter to the test case specification conditional on numpy >= 2.0 ([line 181](https://github.com/maurerv/pyFFTW/blob/59acacb60786785fab157f72e53916c57e65550d/tests/test_pyfftw_numpy_interface.py#L181))

2. Create an appropriately sized array if requested ([line 314](https://github.com/maurerv/pyFFTW/blob/59acacb60786785fab157f72e53916c57e65550d/tests/test_pyfftw_numpy_interface.py#L314))

[...]

PR #421, opened 2025-09-09
Add support for numpy 2.0 `out` argument in pyfftw.interfaces.numpy

maurerv:
## Summary

This PR adds support for the `out` argument introduced in numpy 2.0 to the `pyfftw.interfaces.numpy` module (#409).

## Open Remarks

The changes from the initial discussion primarily target the test suite, making sure that the out argument is not propagated into the respective interface functions.

One issue I stumbled upon in the test cases is the type promotion scheme. `np.fft.fftn` returns complex128 for complex64 inputs whereas `pyfftw.interfaces.numpy_fft.fftn` returns complex64. In order to use the correct output dtype I compute the requested transform twice [see here](https://github.com/maurerv/pyFFTW/blob/a7e828f4eeac79379c04d78de02a31df2ade889f/tests/test_pyfftw_numpy_interface.py#L330).

Is there a way to compute the expected dtype of the transform ahead of time, e.g., have access to fftw_schemes to avoid this redundant computation during testing?


paugier:
I opened a new PR (https://github.com/pyFFTW/pyFFTW/pull/424) with your changes rebased on master (and auto-formatted with Ruff). It should be easier to review for pyFFTW maintainers.

@maurerv Could you please check that I forgot nothing and also if possible reply to my comment https://github.com/pyFFTW/pyFFTW/pull/424/files#r2334736348?

PR #424, opened 2025-09-09
Added out argument to numpy fft interface

### LINKED ISSUE #409: Add support for numpy 2.0 `out` argument in pyfftw.interfaces.numpy

paugier:
Fixes #409
Supersede #421 (same changes - to be checked - but cleaner since rebased on formatted code).

## Summary

This PR adds support for the `out` argument introduced in numpy 2.0 to the `pyfftw.interfaces.numpy` module (#409).

## Open Remarks

The changes from the initial discussion primarily target the test suite, making sure that the out argument is not propagated into the respective interface functions.

One issue I stumbled upon in the test cases is the type promotion scheme. `np.fft.fftn` returns complex128 for complex64 inputs whereas `pyfftw.interfaces.numpy_fft.fftn` returns complex64. In order to use the correct output dtype I compute the requested transform twice [see here](https://github.com/maurerv/pyFFTW/blob/a7e828f4eeac79379c04d78de02a31df2ade889f/tests/test_pyfftw_numpy_interface.py#L330).

Is there a way to compute the expected dtype of the transform ahead of time, e.g., have access to fftw_schemes to avoid this redundant computation during testing?

### DIFF
--- a/pyfftw/interfaces/_utils.py
+++ b/pyfftw/interfaces/_utils.py
@@ -62,6 +62,7 @@ def _Xfftn(
     normalise_idft=True,
     ortho=False,
     real_direction_flag=None,
+    output_array=None,
 ):
     work_with_copy = False
 
@@ -153,7 +154,8 @@ def _Xfftn(
             # the check and the lookup
             FFTW_object = None
 
-    if not cache.is_enabled() or FFTW_object is None:
+    new_object = not cache.is_enabled() or FFTW_object is None
+    if new_object:
         # If we're going to create a new FFTW object and are not
         # working with a copy, then we need to copy the input array to
         # preserve it, otherwise we can't actually  take the transform
@@ -174,23 +176,32 @@ def _Xfftn(
         if cache.is_enabled():
             cache._fftw_cache.insert(FFTW_object, key)
 
-        output_array = FFTW_object(normalise_idft=normalise_idft, ortho=ortho)
-
-    else:
-        orig_output_array = FFTW_object.output_array
-        output_shape = orig_output_array.shape
-        output_dtype = orig_output_array.dtype
-        output_alignment = FFTW_object.output_alignment
-
-        output_array = pyfftw.empty_aligned(
-            output_shape, output_dtype, n=output_alignment
-        )
-
-        FFTW_object(
-            input_array=a,
+    # If output_array is provided FFTW._output_array will be overwritten and the
+    # output of the requested operation on FFTW._output_array is returned. We do
+    # under no circumstances modify this external reference however, as subsequent
+    # operations either trigger the creation of a new FFTW object, or of a new output
+    # array if the output_array argument is None.
+    if new_object:
+        return FFTW_object(
             output_array=output_array,
             normalise_idft=normalise_idft,
             ortho=ortho,
         )
 
-    return output_array
+    # This can be moved above new_object for readability, but comes at a cost of one
+    # additional array instantiation that is not required.
+    orig_output_array = FFTW_object.output_array
+    if output_array is None:
+        output_shape = orig_output_array.shape
+        output_dtype = orig_output_array.dtype
+
+        output_array = pyfftw.empty_aligned(
+            output_shape, output_dtype, n=FFTW_object.output_alignment
+        )
+
+    return FFTW_object(
+        input_array=a,
+        output_array=output_array,
+        normalise_idft=normalise_idft,
+        ortho=ortho,
+    )
--- a/tests/test_pyfftw_numpy_interface.py
+++ b/tests/test_pyfftw_numpy_interface.py
@@ -183,6 +183,23 @@ class InterfacesNumpyFFTTestFFT(unittest.TestCase):
             ((32, 32, 2), {"axis": 1, "norm": "backward"}),
             ((32, 32, 2), {"axis": 1, "norm": "forward"}),
             ((64, 128, 16), {}),
+            ((64, 128, 16), {"out": None}),
+        )
+    elif numpy.__version__ >= "2.0":
+        test_shapes = (
+            ((100,), {}),
+            ((128, 64), {"axis": 0}),
+            ((128, 32), {"axis": -1}),
+            ((59, 100), {}),
+            ((59, 99), {"axis": -1}),
+            ((59, 99), {"axis": 0}),
+            ((32, 32, 4), {"axis": 1}),
+            ((32, 32, 2), {"axis": 1, "norm": "ortho"}),
+            ((32, 32, 2), {"axis": 1, "norm": None}),
+            ((32, 32, 2), {"axis": 1, "norm": "backward"}),
+            ((32, 32, 2), {"axis": 1, "norm": "forward"}),
+            ((64, 128, 16), {}),
+            ((64, 128, 16), {"out": None}),
         )
     else:
         test_shapes = (
@@ -196,6 +213,7 @@ class InterfacesNumpyFFTTestFFT(unittest.TestCase):
             ((32, 32, 2), {"axis": 1, "norm": "ortho"}),
             ((32, 32, 2), {"axis": 1, "norm": None}),
             ((64, 128, 16), {}),
+            ((64, 128, 16), {"out": None}),
         )
 
     # invalid_s_shapes is:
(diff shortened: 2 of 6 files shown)


tests/test_pyfftw_numpy_interface.py
```diff
@@ -183,6 +183,23 @@ class InterfacesNumpyFFTTestFFT(unittest.TestCase):
             ((32, 32, 2), {"axis": 1, "norm": "backward"}),
             ((32, 32, 2), {"axis": 1, "norm": "forward"}),
             ((64, 128, 16), {}),
+            ((64, 128, 16), {"out": None}),
+        )
+    elif numpy.__version__ >= "2.0":
```
paugier: How can this be hit? Moreover, I don't see the difference between the two definitions of `test_shapes` :slightly_smiling_face:

maurerv: The idea for testing was to replace `out` with an appropriate output array if it is explicitly specified in kwarg [here](https://github.com/paugier/pyFFTW/blob/b8902c0c99bc2510d52c56a4a923d066dca3d835/tests/test_pyfftw_numpy_interface.py#L327). Since the cache is disabled, this will create a new FFTW object like before, but using the specified output array. This is followed by the previous test for equivalence with the validator and an additional test for identity with the `out` array in kwarg. 

The `test_shapes` are indeed the same, I think that is on me for not cleaning it up :) At some point, I was thinking to include the `out` argument of the validator in the test case. However, that would have locked testing to numpy >= 2.0 with no real mechanism to prevent its use in <2.0

pyfftw/interfaces/_utils.py
```diff
@@ -174,23 +176,32 @@ def _Xfftn(
         if cache.is_enabled():
             cache._fftw_cache.insert(FFTW_object, key)
 
-        output_array = FFTW_object(normalise_idft=normalise_idft, ortho=ortho)
+    # If output_array is provided FFTW._output_array will be overwritten and the
+    # output of the requested operation on FFTW._output_array is returned. We do
+    # under no circumstances modify this external reference however, as subsequent
+    # operations either trigger the creation of a new FFTW object, or of a new output
+    # array if the output_array argument is None.
+    if new_object:
```
paugier, line 184: @maurerv: I didn't try to fully understand this code. However, I wonder why this `if new_object:` is needed. It seems to me that without this line, the result would be the same.

paugier:
@hgomersall, it would be nice if you can review this PR. For me it's fine, but I don't fully understand the changes in `_utils.py` (I didn't try too much though).

I also wonder if testing is sufficient.

tests/test_pyfftw_numpy_interface.py
paugier: I removed this dead code in eabad86096919d6eeb1bddb01a8d32d6a5a0600f.

paugier:
Should this PR be merged? Or first improved in some way?

paugier:
Python 3.14 will be released in few days so it will be very useful to release a compatible pyfftw version with 3.14 wheels quite soon.

 I tend to think these changes should be included in the release. @hgomersall what is your opinion on that?

hgomersall:
Yeah, I agree. It mostly looks good. Are there any tests that actually exercise the output array arg?

maurerv:
Hi, the output array arg test is integrated into `InterfacesNumpyFFTTestFFT`. To maintain compatibility with other test suites (like the scipy interface), I added [`has_out_kwarg`](https://github.com/paugier/pyFFTW/blob/eabad86096919d6eeb1bddb01a8d32d6a5a0600f/tests/test_pyfftw_numpy_interface.py#L217), analogous to `has_norm_kwarg`. The output array arg is exercised in [test_pyfftw_numpy_interface.py:321](https://github.com/paugier/pyFFTW/blob/eabad86096919d6eeb1bddb01a8d32d6a5a0600f/tests/test_pyfftw_numpy_interface.py#L321).

hgomersall:
Oh I see, the `None` gets overwritten with an actual value in `_validate`.

hgomersall:
It means the `kwargs` aren't really the kwargs in that case.

maurerv:
Exactly!

I think testing the out argument was complicated by the fact that the dtype of input data is not known until [`validate`](https://github.com/paugier/pyFFTW/blob/eabad86096919d6eeb1bddb01a8d32d6a5a0600f/tests/test_pyfftw_numpy_interface.py#L245), and that I did not find how to access the type promotion scheme [fftw_schemes](https://github.com/paugier/pyFFTW/blob/eabad86096919d6eeb1bddb01a8d32d6a5a0600f/pyfftw/pyfftw.pyx#L671) as mentioned in the open remarks of the PR.

So for the time being, adding `out` to `test_shapes` and modifying `kwargs` inside [`_validate`](https://github.com/paugier/pyFFTW/blob/eabad86096919d6eeb1bddb01a8d32d6a5a0600f/tests/test_pyfftw_numpy_interface.py#L255) seemed like the least invasive approach.

hgomersall:
I think a better solution would be to make it explicit, a string like "PLACEHOLDER", to make it clear that it will be updated. Clearly, passing an _actual_ `None` would be the default case. A rust-like enum would be the right option here, but python's enums are pretty horrible.

maurerv:
I see where you are coming from. I think having a dedicated enum would be clearer, or removing `out` from `kwargs` altogether in favor of a boolean, e.g., `test_out_arg`.

paugier:
Could we converge on a nice state for this PR? Python 3.14 is out so it would be nice to release a new version and upload wheels for Python 3.14.

paugier:
I'd need a pyfftw wheel for Python 3.14 on PyPI so I'm going to open a PR preparing a new release.

See also https://discuss.python.org/t/stats-on-python-3-14-support-on-pypi/104319

PR #364, opened 2023-11-03
Fixup/cython errors and readme update

raffienficiaud:
This PR

* fixes the various issues I was experiencing while compiling from sources on my macOS M2, mostly due to `cython` incompatibilities. I believe this supersedes #359 and closes #362 
* provide easier instructions to build locally. This may address/work around #313

PR #426, opened 2025-10-14
PYFFTW_FFTW_PREFIX

### LINKED ISSUE #419: PYFFTW_FFTW_PREFIX and documentation updates
I think it would be useful if pyfftw builds would be sensible to a new environment variable `PYFFTW_FFTW_PREFIX`.

The documentation in the README is quite outdated and should be updated.

For, example, https://docs.python.org/3/distutils/builtdist.html is a dead link. Also we should not mention "running `setup.py`".

Note that the associated PR would supersede https://github.com/pyFFTW/pyFFTW/pull/364

paugier:
Fixes #419
Supersedes https://github.com/pyFFTW/pyFFTW/pull/364

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,10 +135,10 @@ at runtime if asked to perform a transform involving long double precision.
 To build against FFTW libraries at non standard location, [some compilers are
 sensitive to the environment
 variables](https://gcc.gnu.org/onlinedocs/gcc/Environment-Variables.html)
-`CPATH` and `LIBRARY_PATH`. Moreover, you can also use `PYFFTW_INCLUDE` and
-`PYFFTW_LIB_DIR`. If the FFTW libraries still cannot be found, you might also
-need to set the environment variable `CC` to build with the compiler used to
-compile the libraries.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -178,22 +178,14 @@ Install FFTW from [homebrew](http://brew.sh):
 
     brew install fftw
 
-Set temporary environmental variables, such that pyfftw finds fftw:
+Now install pyfftw from source with `PYFFTW_FFTW_PREFIX`:
 
-    export DYLD_LIBRARY_PATH=/usr/local/lib
```
paugier, line 183: I suspect that `DYLD_LIBRARY_PATH` is not actually useful, and this removing was proposed in #364.

README.md
```diff
@@ -123,22 +123,18 @@ That cythonizes the python extension and builds it into a shared library
 which is placed in ``pyfftw/``. The directory can then be treated as a python
 package.
 
-After you've run ``setup.py`` with cython available, you then have a
```
paugier, line 126: It seems that such details are no longer very useful here now that the builder installs build dependencies automatically.

hgomersall (approved):

hgomersall:
It all looks fine. Merge when you want! Thanks!

PR #432, opened 2026-04-14
Add free-threaded CPython 3.14 (cp314t) wheel builds

floens:
Hey, I noticed there were no free-threaded builds on pypi. So, I've looked at the project, and it seems that even though the code (almost) supported free-threaded Python, there were just a few minor things:

- We specify a build matrix. The build tool cibuildwheel seems to build t wheels by default now, but it just wasn't picked up yet.
- Some code around planning wasn't properly thread-safe. We already have a `plan_lock`, so just a bit more locking is done for the planner, wisdom and the global functions. I've used this https://www.fftw.org/fftw3_doc/Thread-safety.html as a guide.

Because I'm not that familiair with the project, I have used AI to assist me. That's reflected in the `Co-Authored-By` commit trailer. I can remove that, depending on what is desired.

Also, because this can mess up the build, I would love someone who worked with the builds to verify that this is correct.

P.S. my apologies, I hit enter a bit too soon, before editing this message :)

### DIFF
--- a/.github/workflows/wheel_tests_and_release.yml
+++ b/.github/workflows/wheel_tests_and_release.yml
@@ -20,7 +20,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-22.04]
-        cibw_python: ["cp311", "cp312", "cp313", "cp314", "pp311"]
+        cibw_python: ["cp311", "cp312", "cp313", "cp314", "cp314t", "pp311"]
         # SciPy and NumPy don't support musllinux
         cibw_libc: ["manylinux"]
         cibw_arch: ["x86_64", "aarch64"]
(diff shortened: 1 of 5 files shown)


hgomersall:
I'm a bit out of the loop on python developments. How does the free-threading stuff work?

floens:
https://py-free-threading.github.io/

https://docs.python.org/3/howto/free-threading-extensions.html

I'm not an expert, but most packages nowadays declare they are free-threaded compatible. Which means that they can run without the GIL. For extensions, they flag compatibility with a `Py_MOD_GIL_NOT_USED` in the module definition. 

By flagging the cython code with a `# cython: freethreading_compatible=True`, it will declare that module as being free-threaded compatible.

If a module doesn't declare it, Python will re-enable the GIL for the entire script. (something like `GIL was re-enabled because module pyfftw does not support disabling the GIL`).

Then for the wheels, they have a `t` suffix in their ABI tag.

hgomersall:
@paugier do you have any thoughts on this?

